Issue #83, opened 2023-02-24
Write down what model correctness is


dslmeinte:
There should be pre-existing text within LIonWeb resources already.

Issue #48, opened 2022-12-09
Allowed characters for names in metamodels


enikao:
Which characters can be used for instances of `Metamodel.qualifiedName` and `NamespacedEntity.simpleName`?

Arguments for limiting to small set (e.g. ASCII):
* Safe

Arguments for big set (e.g. UTF-8):
* We want to name semantic elements according to their domain, which might be Chinese characters for a chinese banking system.

Do we allow spaces in names?

enikao:
On 2023-07-07, we decided to only allow programming language identifiers. More specifically, we allow Java identifiers (https://docs.oracle.com/javase/specs/jls/se20/html/jls-3.html#jls-3.8) with the following modifications:

* We do NOT allow `$` (dollar sign)
* _ReservedKeyword_, _BooleanLiteral_, and _NullLiteral_ (as per Java identifier spec) are allowed identifiers.

Effectively:
* Names MUST NOT start with a number.
* Names MUST NOT be empty.
* Names MUST NOT contain spaces.
* Names CAN use Unicode characters, numbers, and underscore.

**Note: These restrictions only apply to names of Language elements (i.e. M3 concepts/M2 instances).** Any language that uses `INamed` on its own can establish their own constraints.

PR #88, opened 2023-03-04
Write document explaining model correctness

dslmeinte:
Issue: #83


documentation/correctness.adoc
```diff
@@ -0,0 +1,56 @@
...
+* Every key in the `properties`, `children`, or `references` sub objects of the object corresponding to a serialized node is the ID of a feature of the concept pointed to by the `concept` ID value.
+* No additional key-value pairs are present in those sub objects.
+* A key in the `properties` sub object corresponds to ID of a `Property` feature, etc.
+* For each key-value pair in the `properties` sub object: the value has a type matching the `type` specified by the `Property` feature corresponding to the key.
++
+Because the model hasn't been deserialized at this point, incorrectness can't be reported yet as a model with references to nodes.
+Instead, we need another means of referencing a specific location in the serialization, such as an XPath and/or a line/column identification.
+
+
+== Correctness of the deserialized model
```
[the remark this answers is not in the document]
dslmeinte: But for constraints you generally need both the metamodel as well structure-referential soundness, especially the more interesting (and...constraining) constraints involve relations - AKA the links, right?

documentation/correctness.adoc
```diff
@@ -0,0 +1,69 @@
...
+Currently, there's only one version, which is identified as: `"1"` (i.e., a string containing the integer 1).
++
+Note that JSON Schema isn't powerful enough to express more high-level, nor referential constraints on the model serialized as JSON.
+In particular, it's not possible to specify constraints such as that an ID points to an object having that ID.
+An example of such a constraint is the `parent` property of a serialized node.
+
+Hierarchical:: The JSON doesn't violate the following constraints.
++
+* Node IDs are unique (within the serialization).
+* No double serializations (even if the serializations are identical).
```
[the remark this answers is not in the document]
dslmeinte: I took that one out, now.

documentation/correctness.adoc
```diff
@@ -0,0 +1,69 @@
...
+* The `parent` property of a serialized node matches the `id` of the containing node (whenever that containing node is present in the serialization).
+Concretely: if a node `N` is a child of a node `P` then `N.parent` should equal `P`.
+* No double containments, i.e. nodes are not the children of multiple other nodes, nor of themselves.
++
+At this points, links (so: containments and references) don't have to resolve.
+Concretely, this means that IDs that are values of pairs in `children` and `containments` don't need to resolve to a serialized node in the same JSON serialization.
+(They could in another serialization.)
+
```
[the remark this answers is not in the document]
dslmeinte: Done.

documentation/correctness.adoc
```diff
@@ -0,0 +1,69 @@
...
+Meta-referential:: The metamodels specified in the serialization are known and _available_ in the sense that the client('s deserializer) knows how to deserialize models conforming to them.
+That would typically imply the presence of type definitions.
+
+Meta-structural:: The seralization matches the metamodels - in particular:
++
+* Concept identifiers (tuples with a `key` value) are uniquely resolvable to a concept in a metamodel.
+* For every object corresponding to a serialized node:
+** Every key of key-value pairs in the `properties`, `children`, or `references` sub objects is the `key` of a feature of the concept pointed to by the `concept` identifier.
+** No additional key-value pairs are present in those sub objects.
```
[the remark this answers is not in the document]
dslmeinte: That's a hold-over from “version 0” of the serialization format. Fixing that now.

documentation/correctness.adoc
```diff
@@ -0,0 +1,69 @@
...
+* The `parent` property of a serialized node matches the `id` of the containing node (whenever that containing node is present in the serialization).
+Concretely: if a node `N` is a child of a node `P` then `N.parent` should equal `P`.
+* No double containments, i.e. nodes are not the children of multiple other nodes, nor of themselves.
++
+At this points, links (so: containments and references) don't have to resolve.
+Concretely, this means that IDs that are values of pairs in `children` and `containments` don't need to resolve to a serialized node in the same JSON serialization.
+(They could in another serialization.)
+
+Meta-referential:: The metamodels specified in the serialization are known and _available_ in the sense that the client('s deserializer) knows how to deserialize models conforming to them.
+That would typically imply the presence of type definitions.
```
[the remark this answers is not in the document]
dslmeinte: I'm adjusting the text to make it clear the burden is on the deserializing client here. Even if the referred metamodels are available in their JSON format, it's not guaranteed that the serialization chunk can be deserialized using anything other than a “dynamic”/“reflective” mode.

documentation/correctness.adoc
```diff
@@ -0,0 +1,69 @@
...
+
+== Correctness of the deserialized model
+
+[horizontal]
+
+Referential:: All links (on M1/instance-level) resolve.
+Note that the model prior to deserialization doesn't need to be meta-referential or meta-structural!
+
+Constraints:: No constraint violations.
+Constraints can only be checked with full knowledge of all metamodels involved, so meta-referential and meta-structural correctness is necessary for this.
```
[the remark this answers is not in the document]
dslmeinte: I added an idea that needs some more thinking.

documentation/correctness.adoc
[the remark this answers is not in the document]
enikao: Agreed. Thus, I'd move _Structure-referential_ before _Meta-referential_.

resources/correctness-diagrams/levels.svg
enikao, line 1: How about changing the layout a bit: Our main focus in this document is model correctness (layers). So I'd put them as the main part, i.e. left and "leading" the layout of the picture. IMHO, metamodels and (de)serialized versions are secondary.

documentation/correctness.adoc
```diff
@@ -0,0 +1,102 @@
...
+This is typically dependent on the precise JSON Schema validator framework/library/utility used.
+
+=== Hierarchical
+
+The JSON doesn't violate the following constraints.
+
+1. IDs of serialized nodes are unique (within the serialization).
+Violation is to be reported in terms of precise locations in the JSON file of all the serialized nodes with equal ID.
+2. The `parent` property of a serialized node matches the `id` of the containing node (whenever that containing node is present in the serialization chunk).
+Concretely: if a node `N` is a child of a node `P` then `N.parent` should equal `P`.
```
enikao, line 42: `N` can also be an annotation of `P`.

documentation/correctness.adoc
```diff
@@ -0,0 +1,102 @@
...
+Violation is to be reported in terms of the ID of the contain**ing** serialized node, the ID of the contain**ed** serialized node, and the `parent` ID stated by that.
+3. No duplicate containments, i.e. nodes are not the children of multiple other nodes, nor of themselves.
```
enikao, line 44: Extend to annotations

documentation/correctness.adoc
```diff
@@ -0,0 +1,102 @@
...
+Violation is to be reported in terms of the ID of the duplicately-contained serialized node, and the IDs of all serialized nodes that claim to contain it, including the links by which this happens, as well the index in the value of those links.
+
+At this points, links (so: containments and references) don't need to resolve (yet).
```
enikao, line 47: "these points" or "this point"

documentation/correctness.adoc
```diff
@@ -0,0 +1,102 @@
...
+Concretely, this means that IDs that are members of the `children` (for containments) or `targets` (for references) arrays, don't need to resolve to a node in the same JSON serialization.
```
enikao, line 48: extend to annotations

documentation/correctness.adoc
```diff
@@ -0,0 +1,102 @@
...
+**Note** that this is a condition that pertains to the deserializing client, rather than on the serialization chunk.
+
+=== Meta-structural
+
+The seralization matches the metamodels - in particular:
+
+* Concept identifiers (tuples with a `key` value) are uniquely resolvable to a concept in a metamodel.
+Incorrectness is to be reported in terms of the ID of the serialized and the key of the concept which can not be resolved.
+* For every object corresponding to a serialized node:
+** For every entry of the `properties`, `children`, or `references` arrays, the `key` in the `property`, `containment`, resp. `reference` sub object is the key of a feature of the concept pointed to by the `concept` identifier.
```
enikao, line 66: extend to annotations

documentation/correctness.adoc
```diff
@@ -0,0 +1,102 @@
...
+Incorrectness is to be reported in terms of the ID of the serialized node, the feature type (property/containment/reference), the index of the offending entry, and the key which can not be resolved.
+** For every entry of a `properties` array, the value of the `value` property conforms to the `type` specified by the `Property` feature corresponding to the key.
+
+Incorrectness is to be reported in terms of the ID of the serialized node, the feature type (property/containment/reference), the index of the offending entry, and the key which can not be resolved.
+Instead, we need another means of referencing a specific location in the serialization, such as an XPath and/or a line/column identification.
```
enikao, line 71: What does "Instead" refer to?

documentation/correctness.adoc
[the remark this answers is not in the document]
dslmeinte: OK, the thing I'm thinking about is that it might be possible to formulate a meta-DSL for constraints that comes with precise semantics for unresolved links. Often enough, it doesn't really matter that some links can't be resolved, because they aren't navigated over while computing the constraints.

So, in this sense, I don't really see why structure-referential should go before meta-referential, but I'll give it a thunk more.

Stay tuned...

documentation/correctness.adoc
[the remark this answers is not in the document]
dslmeinte: That statement is stronger than the one about just the IDs. As said: I took it out, now.

documentation/correctness.adoc
dslmeinte: Right, I really should implement the annotations in TS as well, so I actually know what's going on there as well.

documentation/correctness.adoc
dslmeinte: I removed that left-over sentence.

documentation/correctness.adoc
```diff
@@ -0,0 +1,123 @@
...
+The JSON doesn't violate the following constraints.
+
+. IDs of serialized nodes are unique (within the serialization).
+Violation is to be reported in terms of precise locations in the JSON file of all the serialized nodes with equal ID.
+. The `parent` property of a serialized node matches the `id` of the containing node (whenever that containing node is present in the serialization chunk).
+Concretely: if a node `N` is a child of a node `P` then `N.parent` should equal `P`.
+Violation is to be reported in terms of the ID of the contain**ing** serialized node, the ID of the contain**ed** serialized node, and the `parent` ID stated by that.
+. No duplicate containments, i.e. nodes are not the children of multiple other nodes, nor of themselves.
+Violation is to be reported in terms of the ID of the duplicately-contained serialized node, and the IDs of all serialized nodes that claim to contain it, including the links by which this happens, as well the index in the value of those links.
+. **TODO**  extend for annotations
```
enikao, line 61: I tend to use `# TODO: ... #` formatting for TODOs, they show up with yellow background.

documentation/correctness.adoc
```diff
@@ -0,0 +1,123 @@
...
+== Model (post-deserialization/in-memory)
+
+=== Structure-referential
+
+All links (on M1/instance-level) resolve.
+Note that this doesn't require any knowledge of the metamodels, as it just involves nodes' IDs.
+This means that the serialization chunk doesn't need to be correct in the meta-referential nor in the meta-structural sense!
+It should be alway possible to deserialize the serialization chunk into "`dynamic`"/"`reflective`" nodes, which could be useful in its own right.
+
+// TODO  describe when and how links should be resolved, etc.
```
enikao, line 102: This line is really commented, i.e. it will not show up in the rendered document at all. Is that desired?

dslmeinte: They are actual TODOs, so they should be rendered. I'll try to find a markup that works - the one you suggested doesn't, sadly.

documentation/correctness.adoc
```diff
@@ -0,0 +1,123 @@
...
+Incorrectness could be partially recovered from by partially deserializing into "`dynamic`"/"`reflective`" nodes.
+
+=== Meta-structural
+
+The seralization matches the metamodels - in particular:
+
+* Concept identifiers (tuples with a `key` value) are uniquely resolvable to a concept in a metamodel.
+Incorrectness is to be reported in terms of the ID of the serialized and the key of the concept which can not be resolved.
+* For every object corresponding to a serialized node:
+** For every entry of the `properties`, `children`, or `references` arrays, the `key` in the `property`, `containment`, resp. `reference` sub object is the key of a feature of the concept pointed to by the `concept` identifier.
```
enikao, line 83: I think this excludes extended features:

```
concept super {
   property x: String
}
concept special extends super {
  property y: Integer
}
```

(simplified chunk format for meta-pointers):

```json5
{
  "id": "mySpecial",
  "concept": "special",
  "properties": [
    {
      "property": "x",
      "value": "hello"
    },
    {
      "property": "y",
      "value": "42"
    }
  ]
}
```

`x` is not "the key of a feature of the concept pointed to by the `concept` identifier".

enikao:
Do we need to incorporate #48?

documentation/correctness.adoc
```diff
@@ -0,0 +1,123 @@
+= Model correctness
+
+We can identify a number of separate criteria for the correctness of a model.
+Failure to comply to one or more of these criteria doesn't imply incorrectness of the model.
```
ftomassetti, line 4: I do not understand this: if the criteria indicate correctness, I would expect that failure to comply should indicate that the model is incorrect _at same level_. It may be still usable, or maybe it could be "partially correct"? Or have some criteria a grade (info, warning) that indicates their severity?

documentation/correctness.adoc
```diff
@@ -0,0 +1,123 @@
...
+It's desirable to be able to recover from "`simple`"/"`obvious`" syntax errors in the JSON.
+
+Incorrectness is to be reported in terms of the location where parsing the JSON text failed, including an indication whether this was recoverable.
+The reporting is typically dependent on the precise JSON parser used.
+
+=== Structural syntactic
+
+The JSON conforms to link:../schemas/serialization.schema.json[the JSON Schema for the serialization format]footnote:[This file should match link:https://github.com/LIonWeb-org/lioncore-typescript/blob/main/schemas/generic-serialization.schema.json[the one in the `lioncore-typescript` repository].].
+The `serializationFormatVersion` property of the root object of the serialization indicates which version to use (or should do so).
+Currently, there's only one version, which is identified as: `"1"` (i.e., a string containing the integer 1).
```
ftomassetti, line 39: I think we recently decided to go for `2023.1` so this may need to be updated

documentation/correctness.adoc
```diff
@@ -0,0 +1,123 @@
...
+At this point, links (so: containments and references) don't need to resolve (yet).
+Concretely, this means that IDs that are members of the `children` (for containments) or `targets` (for references) arrays, don't need to resolve to a node in the same JSON serialization.
+They could -- or even: should -- reside in another serialization.
+That pertains specifically to references to language entities of the LIonCore builtins language, which is assumed to be imported/available implicitly.
+If all links do resolve, we say the serialization chunk is _self-contained_.
+
+=== Meta-referential
+
+All the metamodels specified in the serialization chunk are _made available_ to the client('s deserializer) in a suitable form.
+That would typically imply the presence of types or type definitions in some programming language.
```
ftomassetti, line 72: I do not understand what this means in practice

documentation/correctness.adoc
```diff
@@ -0,0 +1,123 @@
...
+It should be alway possible to deserialize the serialization chunk into "`dynamic`"/"`reflective`" nodes, which could be useful in its own right.
+
+**TODO:** describe when and how links should be resolved, etc.
+
+=== Constraints
+
+No constraint violations.
+Constraints can only be checked with full knowledge of all metamodels involved, so meta-referential, meta-structural correctness is necessary for this.
+Also: the more "`interesting`" constraints typically navigate across references, so structure-referential correctness would also be necessary.
+It should be possible to come up with some kind of "`partial constraint checking`" that deals with unresolved references in a meaningful way -- possibly reporting warnings, rather than outright constraint violations.
```
ftomassetti, line 109: Is this connected to the partial evaluation talk :D ?
This may be a bit too advanced for me to implement

dslmeinte: Yes, it is :) Partial evaluation is actually not that hard to implement - as my talk will show ;)

enikao: Maybe we should remember Federico's mantra: "LionWeb does not want to be innovative, but the boring base to innovate upon."

Especially validations should also be easy to read/understand, as users might not understand why some error is flagged and then try to read/debug the validation implementation. (Some old nightmares from EMF validation / Xtext scoping are resurfacing ...)

documentation/correctness.adoc
```diff
@@ -0,0 +1,123 @@
...
+All links (on M1/instance-level) resolve.
+Note that this doesn't require any knowledge of the metamodels, as it just involves nodes' IDs.
+This means that the serialization chunk doesn't need to be correct in the meta-referential nor in the meta-structural sense!
+It should be alway possible to deserialize the serialization chunk into "`dynamic`"/"`reflective`" nodes, which could be useful in its own right.
+
+**TODO:** describe when and how links should be resolved, etc.
+
+=== Constraints
+
+No constraint violations.
```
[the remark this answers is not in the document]
dslmeinte: This is the correctness criterion. Every sub section starts with a sentence like this.

documentation/correctness.adoc
```diff
@@ -0,0 +1,123 @@
...
+
+
+== Model (post-deserialization/in-memory)
+
+=== Structure-referential
+
+All links (on M1/instance-level) resolve.
+Note that this doesn't require any knowledge of the metamodels, as it just involves nodes' IDs.
+This means that the serialization chunk doesn't need to be correct in the meta-referential nor in the meta-structural sense!
+It should be alway possible to deserialize the serialization chunk into "`dynamic`"/"`reflective`" nodes, which could be useful in its own right.
```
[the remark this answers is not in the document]
dslmeinte: I guess "reflective" would be a little more effective?

documentation/correctness.adoc
dslmeinte: OK, I'll try to expand/improve on it.

documentation/correctness.adoc
dslmeinte: I reworded the 2nd sentence a bit. Instead of a grade, I mentioned "Can an incorrectness be recovered from (and if so: how)?"

documentation/correctness.adoc
dslmeinte: That's true, which is also the reason I need some more time for that proposal.

documentation/correctness.adoc
```diff
@@ -0,0 +1,56 @@
...
++
+* No double containments, i.e. nodes are not the children of multiple other nodes, nor of themselves.
+* The `parent` property of a serialized node matches the `id` of the containing node (whenever that node is present).
+* No (incompatible) double serializations.
+* Node IDs are unique (within the serialization).
++
+Note that links don't have to resolve, so no additional constraints are imposed.
+
+Meta-referential:: The metamodels specified in the serialization are known and _available_ in the sense that the client('s deserializer) knows how to deserialize models conforming to them.
+That might include aspects like type definitions, and metamodel-specific APIs.
```
joswarmer: I do not think this is a validation level,  this is a prerequisite to be able to do validations that need the used language defintions.

documentation/correctness.adoc
```diff
@@ -0,0 +1,133 @@
...
+
+At this point, links (so: containments and references) don't need to resolve (yet).
+Concretely, this means that IDs that are members of the `children` (for containments) or `targets` (for references) arrays, don't need to resolve to a node in the same JSON serialization.
+They could -- or even: should -- reside in another serialization.
+That pertains specifically to references to language entities of the LIonCore builtins language, which is assumed to be imported/available implicitly.
+If all links do resolve, we say the serialization chunk is _self-contained_.
+
+=== Meta-referential
+
+All the metamodels specified in the serialization chunk should be _made available_ to the client('s deserializer) in the form of a runtime representation that the deserializer can reconstruct from the serializatino.
```
joswarmer, line 74: I don't think that remarks on deserialization (like here) are relevant.  In LionWeb we define JSON representations (sofar) for language and models and we should only refer to those.  I think everything mentioned in the  document can be done that way. All the validations heere can be done without actually deserializing anything.
The metamodels should therefore be made available in JSON format.

dslmeinte: That's a good idea: demand that the M2s be available in serialized form.

I think I'd still like to say something about the correctness of deserialized models though, specifically because we also want to say something about reference resolution and constraints, but it wouldn't be bad to separate those aspects more clearly and cleanly.

documentation/correctness.adoc
dslmeinte: Agreed. Will rewrite a bit.

enikao:
I'm very late with this remark, but anyways: What about rewording the document in active voice? Might simplify the language:

Original:

> Incorrectness is to be reported in terms of the location where parsing the JSON text failed, including an indication whether this was recoverable.

Active voice:

> We report incorrectness in terms of the location where we failed parsing the JSON text, including an indicator whether we could recover.

documentation/correctness.adoc
```diff
@@ -0,0 +1,133 @@
...
+Concretely: if a node `N` is a child of a node `P` then `N.parent` should equal `P`.
+Violation is to be reported in terms of the ID of the contain**ing** serialized node, the ID of the contain**ed** serialized node, and the `parent` ID stated by that.
+. No duplicate containments, i.e. nodes are not the children of multiple other nodes, nor of themselves.
+Violation is to be reported in terms of the ID of the duplicately-contained serialized node, and the IDs of all serialized nodes that claim to contain it, including the links by which this happens, as well the index in the value of those links.
+. No containment cycles, i.e. nodes are not containing themselves.
+. **TODO**  extend for annotations
+
+At this point, links (so: containments and references) don't need to resolve (yet).
+Concretely, this means that IDs that are members of the `children` (for containments) or `targets` (for references) arrays, don't need to resolve to a node in the same JSON serialization.
+They could -- or even: should -- reside in another serialization.
```
enikao, line 68: why "should"?

documentation/correctness.adoc
enikao, line 74: typo: "serialization"

documentation/correctness.adoc
```diff
@@ -0,0 +1,133 @@
...
+
+All the metamodels specified in the serialization chunk should be _made available_ to the client('s deserializer) in the form of a runtime representation that the deserializer can reconstruct from the serializatino.
+**Note** that this is a condition that pertains to the deserializing client, rather than on the serialization chunk.
+Incorrectness could be partially recovered from by partially deserializing into "`reflective`" nodes.
+
+=== Meta-structural
+
+The seralization matches the metamodels - in particular:
+
+* Concept identifiers (tuples with a `key` value) are uniquely resolvable to a concept in a metamodel.
```
enikao, line 82: Rename: Concept -> Classifier

documentation/correctness.adoc
```diff
@@ -0,0 +1,133 @@
...
+Incorrectness is to be reported in terms of the ID of the serialized and the key of the concept which can not be resolved.
+* For every object corresponding to a serialized node:
+** For every entry of the `properties`, `children`, or `references` arrays, the `key` in the `property`, `containment`, resp. `reference` sub object is the key of a feature of the concept pointed to by the `concept` identifier.
+Incorrectness is to be reported in terms of the ID of the serialized node, the feature type (property/containment/reference), the index of the offending entry, and the key which can not be resolved.
+** For every entry of a `properties` array, the value of the `value` property conforms to the `type` specified by the `Property` feature corresponding to the key.
+
+Incorrectness is to be reported in terms of the ID of the serialized node, the feature type (property/containment/reference), the index of the offending entry, and the key which can not be resolved.
+"`To resolve a key`" means that we can find a concept or feature in some metamodel with that key, and with (exactly-)matching version.
```
enikao, line 90: I think there's a formatting issue, this doesn't belong to property values

joswarmer:
> I'm very late with this remark, but anyways: What about rewording the document in active voice? Might simplify the language:
> 
> Original:
> 
> > Incorrectness is to be reported in terms of the location where parsing the JSON text failed, including an indication whether this was recoverable.
> 
> Active voice:
> 
> > We report incorrectness in terms of the location where we failed parsing the JSON text, including an indicator whether we could recover.

This document prescribes how things should be done, I think the current way of describing it fits better that the active voice.

documentation/correctness.adoc
```diff
@@ -0,0 +1,133 @@
...
+Incorrectness is to be reported in terms of the location where parsing the JSON text failed, including an indication whether this was recoverable.
+The reporting is typically dependent on the precise JSON parser used.
+In particular, the JSON parser used might recover automatically, essentially ignoring syntax errors without reporting them in any way.
+This could be problematic, e.g. in case of duplicate keys in an object: the JSON parser is most likely to only reproduce the last value for the duplicate key, ignoring the previous ones.
+
+=== Structural syntactic
+
+The JSON conforms to link:../schemas/serialization.schema.json[the JSON Schema for the serialization format]footnote:[This file should match link:https://github.com/LIonWeb-org/lioncore-typescript/blob/main/schemas/generic-serialization.schema.json[the one in the `lioncore-typescript` repository].].
+The `serializationFormatVersion` property of the root object of the serialization indicates which version to use (or should do so).
+Currently, there's only one version, which is identified as a string such as `"2023.1"`.
```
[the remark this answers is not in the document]
dslmeinte: Shortened that paragraph as much as possible.

documentation/correctness.adoc
dslmeinte: Removed

documentation/correctness.adoc
dslmeinte: I don't see the formatting issue.

enikao: Me neither (-:

Maybe I stumbled on "Incorrectness is to be reported  ..." and '"`To resolve a key`" means ...' to be rendered in the same paragraph.

documentation/correctness.adoc
```diff
@@ -0,0 +1,97 @@
...
+
+LionWeb provides a precise notion for model correctness in the form of a multi-leveled specification.
+For each level, it specifies what kinds of incorrectness can arise, how these should be reported as issues, and whether and how these can be meaningfully _recovered_ from.
+This approach provides a degree of resilience: issues arising in one level don't necessarily prohibit determining whether a model exhibits incorrectness on another level, provided an issue can be meaningfully _recovered_ from.
+
+LionWeb provides a JSON serialization format for models: the serialization chunk.
+LionWeb intentionally doesn't prescribe anything about the _programmatic representation_ of models -- i.e.: the runtime or in-memory representation that's the result of deserializing a serialization chunk of a model -- leaving implementors of language-oriented tooling free to choose the representation that's right for them.
+Because of that, this specification is phrased _completely_ in terms of the JSON serialization.
+The first three levels of this specification pertain purely to the JSON that's supposed to be the serialization chunk of a model, and are language(s)-aspecific.
```
enikao, line 30: agnostic?

documentation/correctness.adoc
```diff
@@ -0,0 +1,97 @@
...
+
+== Overview of the levels
+
+We first give an overview of the various levels, and what each pertains to, at first without specifying the corresponding kinds of issues that may arise in that level.
+The list is roughly in order from low-level to language(s)-specific.
+
+[horizontal]
+
+Low-level syntactic::
+Pertains to the character stream that's supposed to be JSON text that's (supposed to be) the serialization chunk of a model.
```
enikao, line 46: Can we simplify this text? This is a summary, so readability is more important than absolute accuracy.

documentation/correctness.adoc
```diff
@@ -0,0 +1,97 @@
...
+This approach provides a degree of resilience: issues arising in one level don't necessarily prohibit determining whether a model exhibits incorrectness on another level, provided an issue can be meaningfully _recovered_ from.
+
+LionWeb provides a JSON serialization format for models: the serialization chunk.
+LionWeb intentionally doesn't prescribe anything about the _programmatic representation_ of models -- i.e.: the runtime or in-memory representation that's the result of deserializing a serialization chunk of a model -- leaving implementors of language-oriented tooling free to choose the representation that's right for them.
+Because of that, this specification is phrased _completely_ in terms of the JSON serialization.
+The first three levels of this specification pertain purely to the JSON that's supposed to be the serialization chunk of a model, and are language(s)-aspecific.
+The next levels _could_ be phrased in terms of a programmatic representation, but the equivalent formulation purely in terms of the serialization format works just as well.
+
+This specification admits some flexibility on the first three, JSON-centric levels.
```
enikao, line 33: Maybe we could borrow the term "well-formed" from XML for these three levels?

enikao: Or maybe that would be "Hierarchical"?

documentation/correctness.adoc
```diff
@@ -0,0 +1,97 @@
...
+
+Meta-structural::
+This means that the serialization chunk conforms to the language(s) it declares to be an instance of.
+An example is that a node must declare values for exactly the features of the classifier it declares.
+
+Referential::
+This pertains to whether references resolve.
+An unresolvable reference doesn't prevent being able to work with the model, but it often prohibits the next level –– Constraints -- from producing meaningful results.
+
+Constraints::
```
enikao, line 64: Does this include or exclude scoping and typing?

documentation/correctness.adoc
```diff
@@ -0,0 +1,97 @@
...
+Hierarchical::
+Pertains to non-local, inter-node constraints on the parsed JSON text.
+An example is that a node that declared a parent node must be contained as a child of that parent node.
+
+Meta-structural::
+This means that the serialization chunk conforms to the language(s) it declares to be an instance of.
+An example is that a node must declare values for exactly the features of the classifier it declares.
+
+Referential::
+This pertains to whether references resolve.
```
enikao, line 61: Only within the chunk or also external?

dslmeinte: Within the chunk indeed.

documentation/correctness.adoc
dslmeinte: Scoping doesn't play so much of a role within a chunk because of explicit (ID-based) references. Of course, the existence of a scoping rule implies the existence of a constraint – e.g., “the field referenced should be a field of the entity referenced by the parent”, so in that sense scoping is included.

Typing is very much included. I think very roughly ~50% of all constraints would be directly “type-informed” anyway.

documentation/correctness.adoc
dslmeinte: “Hierarchical” really pertains to inter-node (implied) references only (so level/aspect 3). I'll add a sentence mentioning “well-formed”.

documentation/correctness-previous.adoc
```diff
@@ -0,0 +1,130 @@
...
+The reporting is typically dependent on the precise JSON parser used.
+In particular, the JSON parser used might recover automatically, essentially ignoring syntax errors without reporting them in any way.
+This could be problematic, e.g. in case of duplicate keys in an object: the JSON parser is most likely to only reproduce the last value for the duplicate key, ignoring the previous ones.
+
+=== Structural syntactic
+
+The JSON conforms to https://github.com/LionWeb-io/specification/blob/main/serialization/serialization.adoc[the specification for the JSON serialization format].
+Note that JSON Schema isn't powerful enough to express more high-level, nor constraints of a referential nature on the serialization chunk.
+
+Incorrectness is to be reported in terms of e.g. a https://github.com/json-path/JsonPath[JsonPath] to the location(s) of failure(s).
```
ftomassetti, line 42: Does this mean that different errors can be reported in a different way? Some specifying the position and some specifying the path?
I think it would be convenient if all of the issues were reported with the position, so that they  can be highlighted in the editor by appropriate plugins. For this it would be useful to have the position as a range of characters (i.e., a starting and an ending point in the file, not a single point).
If some errors specify _also_ a JsonPath, that would be useful

documentation/correctness-previous.adoc
```diff
@@ -0,0 +1,130 @@
...
+
+. If extra properties are present, they can simply be removed or ignored.
+. If properties are missing, they could just be added, provided they have a compliant default value.
+
+=== Hierarchical
+
+The JSON doesn't violate the following constraints.
+
+. IDs of serialized nodes are unique (within the serialization).
+Violation is to be reported in terms of precise locations in the JSON file of all the serialized nodes with equal ID.
```
ftomassetti, line 54: Here do we mean position as in a couple of points indicating line and column for each one or in terms of JSON path?
If we use position, do we start counting from Line 1 and Column 0?

documentation/correctness-previous.adoc
```diff
@@ -0,0 +1,130 @@
...
+. The `parent` property of a serialized node matches the `id` of the containing node (whenever that containing node is present in the serialization chunk).
+Concretely: if a node `N` is a child of a node `P` then `N.parent` should equal `P`.
+Violation is to be reported in terms of the ID of the contain**ing** serialized node, the ID of the contain**ed** serialized node, and the `parent` ID stated by that.
```
ftomassetti, line 57: Also for this I think it would be useful if, in addition to specify the ID of the problematic node, the issues would specify the position at which the error has been identified

documentation/correctness-previous.adoc
```diff
@@ -0,0 +1,130 @@
...
+. No containment cycles, i.e.: nodes don't contain themselves.
+. **TODO**  extend for annotations
+
+At this point, links (so: containments and references) don't need to resolve (yet).
+Concretely, this means that IDs that are members of the `children` (for containments) or `targets` (for references) arrays, don't need to resolve to a node in the same JSON serialization chunk.
+They could very well reside in another serialization.
+That pertains specifically to references to language entities of the LionCore builtins language, which is assumed to be imported/available implicitly.
+If all links do resolve, we say the serialization chunk is _self-contained_.
+
+=== Meta-referential
```
ftomassetti, line 69: If we report category of errors to the users (e.g., syntactic, semantic, lexical, etc.) and if these titles will end up being one of the categories, then I would look for a simpler name, so that a casual user of LionWeb could more easily figure out the meaning

documentation/correctness-previous.adoc
```diff
@@ -0,0 +1,130 @@
...
+All links (on M1/instance-level) resolve.
+Note that this doesn't require any knowledge of the metamodels, as it just involves nodes' IDs.
+This means that the serialization chunk doesn't need to be correct in the meta-referential nor in the meta-structural sense!
+It should be alway possible to deserialize the serialization chunk into "`reflective`" nodes, which could be useful in its own right.
+
+**TODO:** describe when and how links should be resolved, etc.
+
+=== Constraints
+
+No constraint violations.
```
ftomassetti, line 105: What kind of constraints do you have in mind? Are these language-specific constraints (e.g., not having two methods with the same signature, or have at least one constructor, or do not have public methods in a private class)?

documentation/correctness.adoc
```diff
@@ -0,0 +1,106 @@
...
+
+[IMPORTANT]
+====
+We'll talk about _correctness_ from now, and consider _validity_ and _consistency_ to be equivalent concepts.
+====
+
+It makes sense to phrase the answer as a collection of _issues_, instead of giving a hard "`yes/no`" answer.
+An issue is a specific point in the model where a specific kind of incorrectness arises.
+Many (kinds of) issues can be meaningfully recovered from, and recovery can often be obvious to the point it can be automatic.
+The unresolved reference can be left in as-is, and treated as an unknown (but not a _missing_) value with special meaning (i.e., domain-specific semantics).
```
ftomassetti, line 18: Maybe we could state that this may be possible only for advanced consumers of nodes. Some may be able to work only with correct models (for example, most interpreters)

documentation/correctness.adoc
```diff
@@ -0,0 +1,106 @@
...
+The illegal comma should be reported but can subsequently be safely ignored.
+In cases where recovery is not automatic, an issue should be very clear to the user what they can do to resolve it.
+A model must be editable/mutable -- and even usable to some -- even if it has issues.
+
+LionWeb provides a precise notion for model correctness in the form of a multi-leveled specification.
+For each level, it specifies what kinds of incorrectness can arise, how these should be reported as issues, and whether and how these can be meaningfully _recovered_ from.
+This approach provides a degree of resilience: issues arising in one level don't necessarily prohibit determining whether a model exhibits incorrectness on another level, provided an issue can be meaningfully _recovered_ from.
```
ftomassetti, line 25: I am not sure about the term _recovered_. Some issues can be completely recovered from (e.g., missing commas or extra commas in JSON), in the sense that we can operate as if the issue was not there in the first place. Other issues can not be completely _recovered_ from. For example, references that cannot be recovered and are marked as having value unknown are not fully _recovered_ as we cannot get the exact behavior as we would if the reference was solvable. We get some sort-of-working state, that requires extra care from the user (who now needs to handle the unknown values) and some operations cannot be performed at all, because the issue can only be partially recovered.

I would consider using a term different from recovered to indicate all the issues that do not cause a critical stop in loading the model (i.e., if get an XML file instead of a JSON file there is no way to provide even a partial model). Then we could differentiate between issues that can be fully recovered (like extra commas), and issues for which we have a workaround (like using the unknown value)

documentation/correctness.adoc
```diff
@@ -0,0 +1,106 @@
...
+Because of that, this specification is phrased _completely_ in terms of the JSON serialization.
+The first three levels of this specification pertain purely to the JSON that's supposed to be the serialization chunk of a model, and are language(s)-agnostic -- essentially, this is about whether the JSON is _well-formed_ as a serialization chunk in JSON format.
+The next levels _could_ be phrased in terms of a programmatic representation, but the equivalent formulation purely in terms of the serialization format works just as well.
+
+This specification admits some flexibility on the first three, JSON-centric levels.
+That is because the behavior and adaptability of JSON parsers differ from implementation to implementation.
+We simply can't (and don't want to demand) that a LionWeb implementation always produces issues, even if they're recoverable or ignorable anyway.
```
ftomassetti, line 35: I think here the example of JSON issues not reported by some libraries was useful to illustrate what kinds of issues cannot be easily reported on all platforms

documentation/correctness-previous.adoc
dslmeinte: Yup. (It's also taking a bead from our own practice of designing DSLs, I think.)

Gradually “adding on” more information sounds like a good idea. Providing range information should be possible as long as we can use a JSON parser that reports that information per JSON element.

documentation/correctness.adoc
```diff
@@ -0,0 +1,106 @@
...
+
+Low-level syntactic::
+Pertains to the text that's (supposed to be) the JSON serialization chunk of a model.
+
+Structural syntactic::
+Pertains to the syntactic structure of the parsed JSON text.
+It's essentially equivalent to applying a generic JSON Schema.
+
+Hierarchical::
+Pertains to non-local, inter-node constraints on the parsed JSON text.
```
ftomassetti, line 53: Here by non-local we mean that we are considering not a single node but the relations between the nodes _within the given chunk_, correct?
I wonder if someone could interpret non-local to be relative to the relations across different chunks

documentation/correctness.adoc
```diff
@@ -0,0 +1,106 @@
...
+
+Referential::
+This pertains to whether references resolve within the chunk.
+An unresolvable reference doesn't prevent being able to work with the model, but it often prohibits the next level –– Constraints -- from producing meaningful results.
+
+Constraints::
+This pertains to a language(s)-specific notion of correctness.
+An incorrectness would typically either lead to a direct failure in the model's semantics -- i.e.: its execution through interpretation, or generating code and running that --, or to the result of the semantics not making sense in the (context of the) domain.
+This distinction effectively divides this level in two sublevels, but it's up to the language(s) designers to expose those.
+In our experience, a significant part of constraints is "`type-informed`" which means that _type computation/derivation_ is probably an intrinsic part of the constraints aspect of any language.
```
ftomassetti, line 68: Is the type-informed part, one of the two levels you are referring to?

documentation/correctness.adoc
```diff
@@ -0,0 +1,106 @@
...
+Meta-structural::
+This means that the serialization chunk conforms to the language(s) it declares to be an instance of.
+An example is that a node must declare values for exactly the features of the classifier it declares.
+
+Referential::
+This pertains to whether references resolve within the chunk.
+An unresolvable reference doesn't prevent being able to work with the model, but it often prohibits the next level –– Constraints -- from producing meaningful results.
+
+Constraints::
+This pertains to a language(s)-specific notion of correctness.
```
ftomassetti, line 65: I think here an example would be useful

documentation/correctness.adoc
```diff
@@ -0,0 +1,106 @@
...
+Each issue has a _severity_ that is an indication of the recoverability of the issue.
+They are:
+
+[horizontal]
+
+Ignorable::
+
+Fixable::
+
+Stubbable::
```
ftomassetti, line 90: What about calling it `Partially recoverable`?

documentation/correctness.adoc
```diff
@@ -0,0 +1,106 @@
...
+== Overview of the severity
+
+Each issue has a _severity_ that is an indication of the recoverability of the issue.
+They are:
+
+[horizontal]
+
+Ignorable::
+
+Fixable::
```
ftomassetti, line 88: What about calling it `Recoverable`?

dslmeinte: “Fully recoverable”, then, to distinguish from “partially recoverable”.

documentation/correctness.adoc
dslmeinte: No. I rewrote that paragraph to hopefully cause less confusion.

documentation/correctness.adoc
dslmeinte: Your interpretation is correct. I rewrote that text a bit, removing “non-local, inter-node” but explaining it more (hopefully).

I think we have to be relatively lax there as e.g. a parent node doesn't need to be in the serialization chunk.

documentation/correctness.adoc
dslmeinte: Would fully vs. partially recoverable (instead of ignorable vs. stubbable – see other comment) work for you?

documentation/correctness.adoc
dslmeinte: I added text for that.

documentation/correctness-previous.adoc
dslmeinte: Gave an example in the main document.

documentation/correctness-previous.adoc
dslmeinte: I wrote quite some additional text on how to report a location.

documentation/correctness-previous.adoc
dslmeinte: I wrote additional text on how to report locations.

And I would say: (1, 1).

documentation/correctness.adoc
ftomassetti: Yes, thank you

documentation/correctness-previous.adoc
ftomassetti: On this I think ANTLR would start from (1,0).

For example, If I have the text:
```
abc
def
```

`a` would be at Line 1, Column 0. Column 1 would be right after the first character, which seems to make sense to me. I think whatever solution we pick it would be confusing to some, but I would lean towards making the same choice as ANTLR

joswarmer:
**NB:** As a general remark, there are two different things for which we can (and are) define validations:

1. Valid LionWeb Chunks:  these are crucial because our interoperability/interchange is fully based on such chunks.
    Chunks are _by definition_ incomplete.
2. Valid models (language instances): these are validation rules that are defined on complete models. 
    As far as LionWeb is concerned, interoperability/interchange is not affected by this kind of validation errors.
    I don't think we have a precise definition for _complete_ models.
 
We should make a **_very_** clear distinction between these two.

schemas/serialization.schema.json
```diff
@@ -0,0 +1,54 @@
+{
```
enikao, line 1: Doesn't this already exist somewhere?

documentation/correctness.adoc
```diff
@@ -0,0 +1,160 @@
...
+* An identification of the _kind_ of issue.
+(Optionally, this corresponds to an automatic fix.)
+* A _location_: where the issue occurred -- see below.
+* A user-readable message explaining what the issue is, and how it should be fixed.
+Ideally, this message is computed from other information in the issue.
+* The _severity_ of the issue that is an indication of the recoverability of the issue -- see below.
+
+=== Location
+
+An issue is useless if it doesn't tell exactly what “thing” in the model triggered it -- the issue's location.
```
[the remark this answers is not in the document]
dslmeinte: Rewrote it to be less vague.

schemas/serialization.schema.json
dslmeinte: Yeah, it does: I'll see that I can remove it.

documentation/correctness.adoc
```diff
@@ -0,0 +1,160 @@
...
+Hierarchical::
+Pertains to relational constraints within the parsed JSON text.
+An example is that each node ID occurs uniquely.
+Another example is that a node that declares a certain parent node must be contained as a child of that parent node -- provided that parent node is present in the serialization chunk; if it's not, then we can't check that constraint.
+
+Meta-structural::
+This means that the serialization chunk conforms to the language(s) it declares to be an instance of.
+An example is that a node must declare values for exactly the features of the classifier it declares.
+
+Referential::
```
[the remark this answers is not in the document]
dslmeinte: Swapped those around.

dslmeinte:
@markusvoelter @enikao @joswarmer @ftomassetti After a long while I've done some work on this again.

A question to you: Didn't the term “meta-referential” sense at all?

joswarmer:
> @markusvoelter @enikao @joswarmer @ftomassetti After a long while I've done some work on this again.
> 
> A question to you: Didn't the term “meta-referential” sense at all?

To me, not really no.

documentation/correctness/overview.adoc
```diff
@@ -0,0 +1,46 @@
...
+Whether a model is _correct_ —or _valid_, or _consistent_— is an important question.
+
+[IMPORTANT]
+====
+We'll talk about _correctness_ from now, and consider _validity_ and _consistency_ to be equivalent concepts.
+====
+
+Correctness ultimately determines whether a model is useful in the context it exists and is used in.
+Answering that question only with a simple "`yes`" or "`no`" is too simplistic: a model can be incorrect in different ways, and on different levels.
+Having an illegal trailing comma in an array somewhere in the JSON serialization of a model —e.g., `[ 0, 1, 2**,** ]`— should not prevent that serialization chunk from being consumable — i.e.: processable, interchangeable, deserializable, etc.
```
enikao, line 12: if we really want to mention it, I'd move the `i.e.: processable, interchangeable, deserializable, etc.` part to a footnote: This sentence is already hard enough to read without it.

documentation/correctness/overview.adoc
```diff
@@ -0,0 +1,46 @@
...
+Not being able to resolve a reference is a different kind of failure than violating a semantic constraint.
+
+With this document, we aim to provide a useful answer to a narrower question: **When is a _serialization chunk_ correct, and to what extent?**
+We focus purely on serialization chunks, and do not consider programmatic representations of models, e.g. as a result of deserializing serialization chunks.
+(LionWeb intentionally doesn't prescribe anything about the _programmatic representation_ of models — i.e.: the runtime or in-memory representation that's the result of deserializing a serialization chunk of a model — leaving implementors of language-oriented tooling free to choose the representation that's right for them.)
```
enikao, line 17: footnote? `i.e.: the runtime or in-memory representation that's the result of deserializing a serialization chunk of a model`

documentation/correctness/overview.adoc
```diff
@@ -0,0 +1,46 @@
...
+In other words: a post-deserialization, programmatic representation should not add (non-derivable), nor loose information.
+Symmetrically: serialization should not loose or add information either.
+
+* Programmatic representation of a model (post-deserialization) is quite context-specific, and is typically tied to the particular platform/language used — i.e., JVM with Java, Kotlin, Java-/TypeScript, etc.
+This means that the concrete API of the programmatic representation can vary quite a bit, making it more difficult to answer the question in a uniform way.
+
+* The LionWeb serialization JSON format is the cornerstone of the LionWeb specification.
+
+Note that it's entirely up to _consumers_ to assert/validate the correctness of a serialization chunk, and to decide what to do with any incorrectness detected.
+A **consumer** of serialization chunks is a software component accepting serialization chunks for processing (transformation, modification), persistence, etc.
```
enikao, line 32: Is a consumer something different than a client, as in our general architecture description?

documentation/correctness/overview.adoc
```diff
@@ -0,0 +1,46 @@
...
+
+The rest of this document is split up in the following sections:
+
+[horizontal]
+<<issues.adoc#, Issues>>:: How incorrectness is (to be) reported by consumers of serialization chunks.
```
enikao, line 37: "How consumers report incorrectness"?

documentation/correctness/issues.adoc
```diff
@@ -0,0 +1,96 @@
+= Issues
```
enikao, line 1: I _think_ during the discussion of https://github.com/LionWeb-io/lionweb-mps/pull/89 we gravitated towards the term `finding`. Should we use the same here?

documentation/correctness/issues.adoc
```diff
@@ -0,0 +1,96 @@
...
+
+It makes sense to phrase the answer as a collection of _issues_, instead of giving a hard "`yes/no`" answer.
+An **issue** is a specific point in a serialization chunk where a specific kind of incorrectness arises.
+
+Many (kinds of) issues allow a meaningful (full or partial) _recovery_ — i.e.: they can be recovered from.
+Recovery is often obvious enough that it can be performed programmatically — i.e.: automatically.
```
enikao, line 7: I find resulting text like `[...] programmatically — i.e.: automatically. E.g., an illegal [...]` _very_ hard to parse.

documentation/correctness/levels.adoc
```diff
@@ -0,0 +1,110 @@
...
+
+The first three levels pertain purely to the JSON that's supposed to be a serialization chunk, and are language(s)-agnostic — essentially, this is about whether the JSON is _well-formed_ as a serialization chunk in JSON format.
+The remaining two levels require language-awareness, but could still take place entirely within the serialization chunk, although for the final, fifth level —constraints— this is usually cumbersome and better performed on a programmatic representation of the model.
+
+We admit some flexibility on the first three, JSON-centric levels.
+That is because the behavior and adaptability of JSON parsers differ from implementation to implementation.
+E.g., some parsers might simply ignore/skip over trailing commas, or have the last of key-value pairs with duplicate keys "`win`".
+We simply can't (and don't want to demand) that a LionWeb implementation always produces the exact same issues, especially if they are —i.e., have severity:— auto-recoverable.
```
enikao, line 14: `if they are —i.e., have severity:— auto-recoverable.` Syntax error, my brain-parser gave up

documentation/correctness/levels.adoc
```diff
@@ -0,0 +1,110 @@
...
+
+Serialization chunks are by definition allowed to be incomplete.
+It's not important whether references resolve _within_ the serialization chunk: an unresolvable reference doesn't affect interoperability/interchangeability.
+For this reason, we unresolvable references are regarded as an incorrectness purely _within_ the serialization chunk, but one with _stubbable_ severity.
```
enikao, line 18: `rason, we unresolvable`: Superfluous `we`

documentation/correctness/levels.adoc
```diff
@@ -0,0 +1,110 @@
...
+
+== Meta-structural
+
+This level requires knowledge of the languages that a serialization chunk declares to conform to.
+Examples of meta-structural issues are:
+
+* Meta-pointers can't be resolved (within the languages that the serialization chunk declares to conform to).
+* A node must declare values for the features of the classifier it declares.
+
+== Referential
```
enikao, line 79: This level is missing in the overview above.

documentation/correctness/levels.adoc
```diff
@@ -0,0 +1,110 @@
...
+It's up to the language(s) designers to make that distinction (whenever it exists) clear to the language's users.
+
+We give an example of a language-specific constraint.
+Consider a language with core concepts _tables_, _columns_ within those, and SQL-like _queries_.
+These queries _reference_ columns within tables, e.g., in the form: `<table>.<column>`.
+A constraint for any reference would be: "`in a reference to a column of a referenced table, the column referenced must be a column of the referenced table`".
+
+Another example would be that names should appear uniquely within bounded contexts.
+
+In our experience (and independent of/orthogonal to the two sublevels), a significant part of constraints is "`type-informed`" which means that _type computation/derivation_ is typically an intrinsic part of the constraints aspect of any language.
```
enikao, line 97: Which sublevels?

PR #133, opened 2023-05-30
added basic asciidoctor workflow

enikao:
coordinated with https://github.com/LIonWeb-org/.github/pull/7


.github/workflows/asciidoctor-ghpages.yml
```diff
@@ -0,0 +1,33 @@
+name: GitHub Pages Publish
+
+# Controls when the action will run. Triggers the workflow on push or pull request
+# events but only for the master branch
+on:
+  push:
+    branches: [ master, main ]
```
ftomassetti, line 7: I think we just use `main`

ftomassetti (approved):
It looks good to me, but I am not very familiar with asciidoc or this particular github actions. If it works, I am happy with it :)

.github/workflows/asciidoctor-ghpages.yml
enikao: This is taken verbatim from https://github.com/manoelcampos/asciidoctor-ghpages-action/blob/main/.github/workflows/asciidoctor-ghpages.yml

Issue #91, opened 2023-03-06
Requirements on metamodel keys


enikao:
Both `NamespacedElement.key` and `Metamodel.key` MUST adhere to LIonWeb id format (#31).

`NamespacedElement.key` MUST be unique within its metamodel.

`Metamodel.key` SHOULD be globally unique, and MUST be unique within an id-space (#31).
For approximate global uniqueness, we SHOULD adopt Java's package naming scheme, based on [domain names](https://en.wikipedia.org/wiki/Domain_name).
As we don't allow dots (`.`) in ids, we SHOULD use dashes (`-`) instead.
Dashes are preferable to underscores (`_`) because of better Java compatibility: In Java package names, underscores are allowed, but dashes are not. The same applies to most other programming languages, as they need to separate names from _subtract_ expressions. Thus, we can easily establish similar names for metamodels and packages by substituting dashes with dots.

Example: Given domain `subdomain.example.com`, metamodel keys might be `com-example-subdomain-myDepartment-group123-entity_model` (with package name `com.example.subdomain.mydepartment.group123.entity_model`) and `com-example-subdomain-great_ultimate_universal_unified_dataModel` (with package name `com.example.subdomain.great_ultimate_universal_unified_datamodel`).

enikao:
Agreed in meeting on 2023-06-09

enikao:
For reference: The agreed ids and keys are listed in [metamodel docs](https://lionweb-org.github.io/organization/lioncore/metametamodel/metametamodel.html#_pre_defined_ids_and_keys)

PR #140, opened 2023-06-09
adjusted M3 docs

enikao:
* Moved serialization and M3 docs to separate directories
* Reordered sections in M3 doc
* Updated pre-defined ids and keys in M3 / builtins
* Added builtin `INamed` and use it in M3
* Added `Concept.partition` property
* Added reference definitions of M3 and builtins as JSON
* Updated serialization format class diagram


lioncore/metametamodel/lioncore.json
```diff
@@ -0,0 +1,2196 @@
...
+  "serializationFormatVersion": "1",
+  "languages": [
+    {
+      "key": "LIonCore-M3",
+      "version": "1"
+    }
+  ],
+  "nodes": [
+    {
+      "id": "-id-LIonCore-M3",
```
joswarmer, line 11: All id values start with a minus, is this really the intention? Or is something missing?

lioncore/metametamodel/lioncore.json
```diff
@@ -0,0 +1,2196 @@
...
+      "concept": {
+        "language": "LIonCore-M3",
+        "version": "1",
+        "key": "Language"
+      },
+      "properties": [
+        {
+          "property": {
+            "language": "LIonCore-M3",
```
joswarmer, line 20: Did we not agree on using underscores instead of minusses? Or is that at some other place?

joswarmer: I checked, and in previous versions we user underscores. I prefer underscores, because they are allowed in almost every environment within identifiers, while minusses are not.

lioncore/metametamodel/lioncore.json
```diff
@@ -0,0 +1,2196 @@
...
+            "version": "1",
+            "key": "LIonCore-builtins-INamed-name"
+          },
+          "value": "LIonCore.M3"
```
joswarmer, line 24: And here we use a dot. I am confused as to where (and why) we use dot, minus and underscore.

lioncore/metametamodel/lioncore.json
enikao: I deliberately used different _ids_ and _keys_, so it's clear everywhere which one we refer to.
In principle the _id_ should be arbitrary (as long as they resolve). As I wrote this file by hand, I used kind of meaningful ids.

lioncore/metametamodel/lioncore.json
enikao: In #91 we decided to change _keys_ to use minuses (rationale in the issue).

lioncore/metametamodel/lioncore.json
enikao: This is the name of the language. It has no technical meaning, so could also be "LIonWeb LIonCore M3 Metamodel" or anything else.
As naming is hard, I chose this one without any imagination. And made it different from _key_ and _id_ for aforementioned reason.

joswarmer: I agree naming is hard indeed !

lioncore/metametamodel/lioncore.json
```diff
@@ -0,0 +1,2196 @@
...
+            "version": "1",
+            "key": "Language-elements"
+          },
+          "children": [
+            "-id-Concept",
+            "-id-ConceptInterface",
+            "-id-Containment",
+            "-id-DataType",
+            "-id-Enumeration",
+            "-id-EnumerationLiteral",
```
joswarmer, line 56: EnumerationLiteral is a containment within Enumeration, not within Language (see question on slack)

lioncore/metametamodel/lioncore.json
```diff
@@ -0,0 +1,2196 @@
...
+            "-id-Feature",
```
joswarmer, line 57: Feature is contained in FeaturesContainer, nit in Language.

lioncore/metametamodel/lioncore.json
```diff
@@ -0,0 +1,2180 @@
...
+        {
+          "reference": {
+            "language": "LIonCore-M3",
+            "version": "1",
+            "key": "Concept-implements"
+          },
+          "targets": []
+        }
+      ],
+      "parent": "LIonCore-M3"
```
ftomassetti, line 2042: This should be "-id-LIonCore-M3", I think

lioncore/metametamodel/lioncore.json
```diff
@@ -0,0 +1,2180 @@
...
+        {
+          "property": {
+            "language": "LIonCore-M3",
+            "version": "1",
+            "key": "LIonCore-builtins-INamed-name"
+          },
+          "value": "NamespacedEntity"
+        },
+        {
+          "property": {
```
ftomassetti, line 1747: Is Key a property of NamespacedEntity? I was not expecting this

In the Java implementation I have an interface called HasKey, but we discussed the fact that it should be just an implementation detail, so we do not have a common point where to define the key.

I think that NamespacedEntity could not play this role, as the Language has a key but it is not a NamespacedEntity.

I think that we have two options:
- either we use two key properties (one declared under Language and one under LanguageElement)
- or we introduce a common ancestore for Language and LanguageElement which defines the key

enikao: Currently option 1 is implemented here. Option 2 would be `HasKey`. As you say, we decided against `HasKey`, which would leave Option 1.
Conceptually, I agree `key` is a bit out of place within `NamespacedEntity` -- `key` is a language thing, namespaces could be used outside this context.
We could rediscuss this tomorrow?

lioncore/metametamodel/lioncore.json
enikao: as discussed on slack: correct as-is.

lioncore/metametamodel/lioncore.json
enikao: as discussed on slack: correct as-is.

lioncore/metametamodel/builtins.json
```diff
@@ -0,0 +1,360 @@
...
+      "id": "LIonCore-builtins",
+      "concept": {
+        "language": "LIonCore-M3",
+        "version": "1",
+        "key": "Language"
+      },
+      "properties": [
+        {
+          "property": {
+            "language": "LIonCore-M3",
```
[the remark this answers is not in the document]
ftomassetti: Yes, indeed

lioncore/metametamodel/lioncore.json
ftomassetti: Sure, makes sense

lioncore/metametamodel/builtins.json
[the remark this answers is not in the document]
enikao: already applied (-:

lioncore/metametamodel/builtins.json
```diff
@@ -0,0 +1,360 @@
...
+          },
+          "value": "String"
+        },
+        {
+          "property": {
+            "language": "LIonCore-M3",
+            "version": "1",
+            "key": "IKeyed-key"
+          },
+          "value": "LIonCore-builtins-String"
```
ftomassetti, line 94: For M3 ids are different from keys, while for the builtins they match. For consistency I would make them different, following the same schema used in M3

lioncore/metametamodel/lioncore.json
```diff
@@ -0,0 +1,2066 @@
...
+        {
+          "reference": {
+            "language": "LIonCore-M3",
+            "version": "1",
+            "key": "Property-type"
+          },
+          "targets": [
+            {
+              "resolveInfo": "Boolean",
+              "reference": "LIonCore-M3-Boolean"
```
ftomassetti, line 205: I think this should be `LIonCore-builtins-Boolean`

lioncore/metametamodel/lioncore.json
```diff
@@ -0,0 +1,2066 @@
...
+        {
+          "reference": {
+            "language": "LIonCore-M3",
+            "version": "1",
+            "key": "Concept-implements"
+          },
+          "targets": [
+            {
+              "resolveInfo": "IKeyed",
+              "reference": "IKeyed"
```
ftomassetti, line 1411: I think this should be `-id-IKeyed`

lioncore/metametamodel/lioncore.json
```diff
@@ -0,0 +1,2066 @@
...
+        {
+          "reference": {
+            "language": "LIonCore-M3",
+            "version": "1",
+            "key": "Property-type"
+          },
+          "targets": [
+            {
+              "resolveInfo": "String",
+              "reference": "LIonCore-M3-String"
```
ftomassetti, line 1580: I think this should be `LIonCore-builtins-String`

lioncore/metametamodel/lioncore.json
```diff
@@ -0,0 +1,2066 @@
...
+            "-id-IKeyed-key"
+          ]
+        }
+      ],
+      "references": [
+        {
+          "reference": {
+            "language": "LIonCore-M3",
+            "version": "1",
+            "key": "Concept-extends"
```
ftomassetti, line 1705: IKeyed is an interface, therefore it is using `ConceptInterface-extends`

lioncore/metametamodel/lioncore.json
```diff
@@ -0,0 +1,2066 @@
...
+      "children": [
+        {
+          "containment": {
+            "language": "LIonCore-M3",
+            "version": "1",
+            "key": "Classifier-features"
+          },
+          "children": [
+            "-id-Concept-abstract",
+            "-id-Concept-extends",
```
ftomassetti, line 131: It should also have the partition feature

lioncore/metametamodel/lioncore.json
```diff
@@ -0,0 +1,2066 @@
...
+              "resolveInfo": "Language",
+              "reference": "-id-Language"
+            }
+          ]
+        }
+      ],
+      "parent": "-id-Language"
+    },
+    {
+      "id": "-id-Language-elements",
```
ftomassetti, line 1478: Should this become entities?

lioncore/metametamodel/builtins.json
enikao: This is something I'm not sure about, see https://lionweb-org.github.io/organization/lioncore/metametamodel/metametamodel.html#_ids_of_built_in_elements

What do you think?

lioncore/metametamodel/lioncore.json
enikao: fixed

lioncore/metametamodel/lioncore.json
enikao: fixed

lioncore/metametamodel/lioncore.json
enikao: fixed

lioncore/metametamodel/lioncore.json
enikao: fixed (removed `Concept-extends`, renamed `Concept-implements` to `ConceptInterface-extends`)

lioncore/metametamodel/lioncore.json
enikao: fixed

lioncore/metametamodel/builtins.json
ftomassetti: I understand that both keys and IDs must be stable, but I do not understand why they need to be identical. However I have just a very tiny preference for them being difficult for consistency with M3 and to catch bugs more easily, but definitely not a strong opinion about this

enikao: My line of thought: Node ids must only be consistent within one _context_. A context might be one json file, or one repository. Once we move from one context to another (e.g. importing a json file into a repository), all ids might change. That's ok, as long as they change consistently.

(If we later wanted to import another json file that refers to the first one, our repository has to keep the ids of the first one around, but that import issue is not relevant here.)

Keys must be consistent _accross contexts_. Otherwise we always have to ship the M2 along with the M1. (We already decided that.)
This also applies to the keys in `builtins`.

The `builtins` model is special in the sense that we _never_ ship it between contexts -- we always assume it's there.
Thus, we also need stable ids in `builtins`, otherwise we cannot refer to them. But we have to refer to them in every M2.

So yes, we can keep keys and ids different, as long as both are stable.

For `lioncore.json`, the ids (in form `-id-xxx`) don't really matter, as they are gone after import to our repository. We never want to refer to these nodes from the same meta-level (nobody can add/remove things at M3 level). References from other meta-levels use keys.
If we wanted to refer to M3 elements as regular models (e.g. in a documentation language: "Please create a new ->M3.Concept"), these elements are "just" regular nodes, so we refer to them by which id they have in the local context.

For `builtins.json`, the ids _do_ matter. That's why I think the consistency between `lioncore` and `builtins` is not that relevant.

I agree debugging is easier if id differs from key. On the other hand, implementations have to keep two sets of "magic" strings around.

Long story short: I have a slight preference for keeping key == id for `builtins`, but also not a strong opinion.

enikao: Decision for now: Keep the same. Revisit once we have experience or discover issues.

PR #151, opened 2023-07-28
Restructured documentation folders

enikao:


metametamodel/metametamodel.adoc
```diff
@@ -577,7 +577,7 @@ The language itself has key `LIonCore-M3`.{fn-org91}
 |<<Reference>> |Concept |Reference
 |===
 
-==== Ids of Built-in elements
+==== Ids of built-in elements
```
ftomassetti, line 580: It seems that in other files you moved to capitalized titles. Does it depend on the level of the header?

ftomassetti:
It seems all in order, I just added a small comment. By the way, I am now trying to use AsciiDoc on my own project after seeing what you built with it!

metametamodel/metametamodel.adoc
enikao: I tried to normalize to
* Document titles are all capitalized
* Headers are sentence capitalized

PR #155, opened 2023-08-04
Add annotations to M3

enikao:


dslmeinte (approved):
Looks good to me, although I'm not yet (re-)qualified to say anything definitive about the annotations part.

metametamodel/builtins.json
```diff
@@ -67,6 +67,7 @@
           "targets": []
         }
       ],
+      "annotations": [],
```
ftomassetti, line 70: I am afraid I will need to manually fix a ton of files in the lioncore-java repo :(

ftomassetti: I wonder if we should revisit the decision of necessarily having to write empty lists

ftomassetti: Some tests relies on JSON files in the resources and I would need to update those files. Those files have been obtained from various sources, and I do not have the code to generate all of them. Previously performing changes on those files have been quite time consuming

dslmeinte: It would be interesting to see how we could generate those files after all.

ftomassetti: I am referring to the files that I used in tests of the Java implementation, like this one: https://github.com/LIonWeb-org/lioncore-java/blob/master/core/src/test/resources/serialization/TestLang-language.json
Given they are files used in tests, I do not think I should generate them

metametamodel/metametamodel.adoc
```diff
@@ -139,6 +141,161 @@ A Concept can have any number of <<Classifier.features, `features`>>, given it i
...
+==== Annotation
+An Annotation is a limited[*] additional, orthogonal information attached to potentially any node, sharing the node’s lifecycle.
+The annotated node (or its concept) does not need to know about the annotation.{fn-org13}
+
+.Example
+We design a fancy documentation language, and can annotate any other node with such documentation.
+
+.EMF & MPS equivalent
+An Annotation is equivalent to MPS' `NodeAttribute`.
+In EMF, an `Adapter` can be used for similar purposes.
```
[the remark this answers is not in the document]
enikao: Good catch! I'll look into it.
From vague memory, they only work on `.ecore` files, i.e. metamodels.

joswarmer (approved):

Issue #157, opened 2023-08-11
Maintain order in serialization of each containment / reference / annotation


enikao:
For each node, we serialize these lists:
* properties
* children
* references
* annotations

We exclude properties from this issue, as we don't allow multiple values per property.
Thus, they fall into scope of #156.

What requirements do we have on the order of children / references / annotations?

Assume this language:

```
concept Person {
  containment addresses: Address[0..*]
  containment cars: Vehicle[0..*]
  reference siblings: Person[0..*]
  reference friends: Person[0..*]
}

annotation Tag {
  multiple = true
  annotates = Person

  property name: String
}

annotation Comment {
  multiple = false
  annotates = Node

  property text: String
}
```

and this instance:

```
Person P {
  addresses: [
    Address { mainStreet }
    Address { mountainWay }
  ]

  cars: [
    Auto { Lotus }
    Motorcycle { Honda }
  ]

  siblings: [
    --> C
    --> D
  ]

  friends: [
    --> D
    --> E
  ]

  annotations: [
    Tag { Adult }
    Comment { Known to know stuff }
    Tag { Blonde }
  ]
}
```

For each of
* `mainStreet` / `mountainWay`
* `Lotus` / `Honda`
* `--> C` / `--> D`
* `--> D` / `--> E`
* `Adult` / `Known to know stuff` / `Blonde`
We have the following options:

## Option A: Order is undefined
The order of serialization is undefined. Any LIonWeb client is free to chose their order, and may or may not keep the order when serializing back the (unchanged) model.

Pro:
* Very simple to implement
* Can be supported by most existing systems

Con:
* Very surprising to the user: They create the model with order  `mainStreet` / `mountainWay`, and after reloading it might show `mountainWay` / `mainStreet`

## Option B: Order must be maintained
The order of serialization is relevant. Any LIonWeb client needs to keep the order as-is, unless the order is explicitly changed (e.g. the user reorders the entries).

Pro:
* In line with user expectations

Con (copied from #4):
* For automatic merging / diffing, explicitly knowing the order doesn't matter is very useful
* Collaborative modeling / distributed data structures can handle unordered collections a lot easier

## Option C: Order by defined criteria
For example, entries are ordered by their ID, or some other criteria.

Pro:
* Order inherent to the data, so serialization order is irrelevant

Con:
* Finding sensible criteria is hard
* Very un-intuitive for users: The add a new entry somewhere, and it might jump somewhere else based on the ordering criteria

[...]

Issue #153, opened 2023-08-04
Details on builtin language


enikao:
# Are builtins part of regular models or always assumed present?

When we export a model to a file (as we are doing in several scenarios) do we export the builtin primitives as well? Or do we leave them out and is any tool reading the file assumed to know them?

Pro part of export:
* Self-contained JSON file
  BUT: It's hard to make sense of the json content without specific LW knowledge

Pro always present:
* Part of installation with special, stable ids
* Versioned

**Decision: We assume builtins are always present.**

# Do we declare builtins as language dependency?

If my language _A_ uses anything of language _B_, _A_ needs to _depend on_ _B_.
If _A_ uses anything of built-in, does it need to _depend on_ _Builtin_?

Pro:
* Uniform handling of all language dependencies

Con:
* Every language (transitively) depends on _Builtin_. In #50, we decided we _do not_ need to declare transitive language dependencies.
* Any new language would need to declare the dependency before we can set e.g. a builtin property type

**Decision: Don't need to declare**

# Do we declare builtins as used language?
Assume a JSON file that contains the M2 of language _A_. _A_ refers to builtin entities like `Node` or `String`.
In the _used languages_ section (#76), do we mention _builtins_?

Pro:
* As _builtins_ are used in M3, any M2 will probably instantiate _builtins_ entities.
  Thus, _builtins_ is a _used language_.
  NOTE: M1 JSON files might _not_ have _builtins_ as _used language_, e.g. if they don't contain any instance of `INamed`.
   But, if they do (e.g. my language defines `Person implements INamed`, and my M1 contains an instance of `Person`), we _do_ use _builtins_.
* If my language B does _not_ depend on builtin types, because I declare my `SpecialSting` and `SpecialNumber` types, I _don't_ want to import _builtins_.

Con:
* It's impossible to define a language without using _builtins_, so we can assume it to always be used.

**Decision: _builtins_ are handled like any other language on serialization level, i.e. MUST be listed as _used language_ if instantiated anywhere in the serialization chunk.**

PR #158, opened 2023-08-31
Adjust docs to ordering decisions

enikao:
Mainly #156 / #157 
Also #153, #155

### DIFF
--- a/metametamodel/metametamodel.adoc
+++ b/metametamodel/metametamodel.adoc
@@ -98,12 +98,12 @@ Language elements contained in a Language are allowed to refer to these other La
 * within declared dependencies
 * within _transitive_ dependencies.
 +
-Example: if Language B depends on Language A, and Language C depends on B, then Language C can also refer to members of Language A without explicitly declaring a dependency.
+Example: if Language _B_ depends on Language _A_, and Language _C_ depends on _B_, then Language _C_ can also refer to members of Language _A_ without explicitly declaring a dependency.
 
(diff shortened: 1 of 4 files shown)


metametamodel/metametamodel.adoc
```diff
@@ -139,11 +139,12 @@ A Concept `implements` zero or more <<ConceptInterface, ConceptInterfaces>>.
 A Concept can have any number of <<Classifier.features, `features`>>, given it is a <<Classifier>>.
 
 .Constraints
-TBD
+A Concept MUST NOT extend itself, or form circles via `extends`.
 
 [[Annotation]]
 ==== Annotation
-An Annotation is a limited[*] additional, orthogonal information attached to potentially any node, sharing the node’s lifecycle.
+An Annotation is a limitedfootnote:["limited" because the more complex the annotation is, the more likely it should not be part of the annotated node's lifecycle -- We might want to reuse the complex annotation somewhere else.
```
dslmeinte, line 146: Proposal: "[...]is an additional piece of information[...]", and then explain limited + orthogonal in a separate sentence. That way, you probably don't even need the footnote.

dslmeinte (changes requested):

metametamodel/metametamodel.adoc
enikao: Good idea, what do you think of this attempt?

dslmeinte (approved):

PR #161, opened 2023-09-01
Fix wrong type in canonical LIonCore JSON serialization chunk

dslmeinte:

PR #162, opened 2023-09-08
Correct key of Annotation-annotates type

ftomassetti:


dslmeinte:
Is same as PR #161 , right?

ftomassetti:
Sorry for the duplication!

PR #166, opened 2023-09-18
SVG LionWeb logo

enikao:


resources/LangDev Talk.pptx
dslmeinte, line 1: Should this file be here, in this PR/branch?

resources/logo-path.svg
dslmeinte, line 1: What's the difference between path and non-path?

resources/LangDev Talk.pptx
enikao: Not really ... This was what I started with, and created lots of other stuff for it. I'll take it out.

resources/logo-path.svg
enikao: non-path contains the text as text, `path` are pure shapes. The latter is safer for display, as it doesn't rely on the font being available.

dslmeinte: Maybe mention this in a `README.md` in that directory?

Issue #94, opened 2023-03-06
Provide id mapping API


enikao:
A client could ask the repository (or another processor) to keep a mapping from an external node id to the internal node id.

We might also have parallel APIs to the regular ones (bulk access, delta access, ...) based on the external node ids.
Example: Parallel to "give me the bulk subtree under node id 234", we could ask "give me the bulk subtree under node id my.packagename.VeryGreatClass.specialMethod(java.lang.String)"

Use case: The external model is a parse-based AST, and we use fully qualified names as ids. After importing the ASTs, we want to re-import parts of the AST, without losing references between existing parts and new parts.

The repository or processor might achieve the mapping by storing an internal property for the external node id, or having a unique mapping function (e.g. base64url, if the repository supports full LIonWeb id range).

enikao:
Deciding whether external ids are mapped as part of Bulk API (#127 / #164) is a trade-off. Let's look at different alternatives.

# Assumptions:
* ids always adhere to #31 
* we always need a "get unused ids" request in the API, if a client wants to be 100 % sure to get an unused id.

#Term: "case id"
A client wants to represent external entities in the model; these entities are identified by their "case id".
The "case id" is _NOT_ the same as a node id (#31).

Example: parser client
* In the first run, it parses the Java class `com.example.Car`. The parser uses the FQN as "case id" and creates a representation of `Car` in the model.
* In the second run, the parser parses the Java class `com.example.Car` again, and updates the representation in the model (e.g. adds new fields).
* In the third run, the parser parsers another Java class `com.somethingelse.Person` and finds an statement `import com.example.Car;`. The parser now needs to find the node representing `Car` in the model.

# A: Server restricts id-space
A server can restrict accepted ids to a subset of #31, e.g. only integers.

## Server:
* can use single index (in database sense) for
  * guarantee id uniqueness
  * efficient access to nodes
* potentially simple bookkeeping on free ids (e.g. only store "highest already used integer")

## Client
* Needs to ask for available ids (and keep them around, as it probably doesn't want to ask the server again for every new node)
* Needs to map its own "case id" to server-provided ids
  * Mapping needs to be stable when re-creating from original source
  * (similar to the separation of M2 `key` vs. `id` #91)
  * same as MPS java stubs: uses FQNs as internal id (`ForeignId`)

Pro:
* Centralized owner of ids simplifies the mental model of the system

Con:
[...]

PR #167, opened 2023-09-18
bulk API spec

enikao:


enikao:
Bulk API depends on solving #94

bulk/repo-access-api.adoc
```diff
@@ -68,6 +68,13 @@ NOTE: We might add advanced filtering capabilities in the future, or introduce a
 
 If the repository supports <<trash>>, it MUST retrieve nodes from _trash_ if the client asks for them.
 
+[horizontal]
```
enikao, line 71: This came up during the last discussion (although not with that name). The idea was that _deleting_ a node should be an explicit action, not merely implicitly by not mentioning it anywhere. Thus we need a place to keep such nodes.

bulk/repo-access-api.adoc
```diff
@@ -119,6 +126,9 @@ We always process one node in its entirety, i.e. we cannot update parts of the n
 A node id referenced as parent, containment, reference, or annotation _can_ be mentioned in the same request, but _can_ be omitted if a node with that id already exists in the repository.
 This way, we can move subtrees and add arbitrary references without sending unchanged nodes.
 
+[horizontal]
```
enikao, line 129: Not as parent, because that would break parent/containment symmetry (and this node would be contained in a non-existing node, thus not be part of a model tree). The others to some degree yes, as we support invalid models.
I can definitely see that for references. Would need to think about the implications of dead containments/annotations.

bulk/repo-access-api.adoc
```diff
@@ -154,6 +164,14 @@ The repository SHOULD NOT validate the "meta-semantics" of the sent nodes, e.g.
 Rationale: We want to support storing invalid models, so the user does not lose data.
 Also, we assume no knowledge of the metamodel.
 
+[horizontal]
```
enikao, line 167: My approach would be to separate model _storage_ (what we discuss here) from model _validation_. For storage, I'd aim at the least restrictions possible.

bulk/repo-access-api.adoc
```diff
@@ -231,6 +249,9 @@ link:https://www.rfc-editor.org/rfc/rfc9110.html#name-delete[RFC 9110: HTTP Sema
 > An origin server SHOULD NOT rely on private agreements to receive content, since participants in HTTP communication are often unaware of intermediaries along the request chain.
 
 Thus, this example sends the node ids via URL query.
+
+[horizonmtal]
```
enikao: I agree, but that's what the RFC suggests.

bulk/repo-access-api.adoc
```diff
@@ -1,155 +1,335 @@
...
+After completing this call, all sent nodes MUST have exactly the sent contents (except `parent`, see below #TODO correct?#) in the repository.
+We must send containments/annotations from the parent's view, because we need to know the containment and index of the contained node/annotation within its parent.
+
+If we move a contained/annotation node `C` from its previous parent `A` to its new parent `B`, we MUST send `B`, and MAY omit `A`.
+This means we can have implicit changes in `A`.
+
+The whole call fails, without any changes to the repository, if it would lead to a malformed model (i.e. cycles in containment hierarchy, same node contained in different parents, parent and corresponding containment/annotation don't match #TODO: Add reference to model validation#).
+
+The repository SHOULD NOT #TODO: or MUST NOT?# validate the "meta-semantics" of the sent nodes, e.g. whether a Concept instance without parent is a partition, a feature instance can be part of its node instance, or a node mentioned as annotation is an Annotation instance.
+Rationale: We want to support storing invalid models, so the user does not lose data.
```
enikao, line 174: Jos:: The repository should work without metamodel knmowledge, but some kind of correctness will be needed,
otherwise many clients won't work.
We should probably cross-reference with the validation document to see which rules are to be maintained by the repository.
Jos:: If the metamodel is know n, the repository can check many extra things.
That _could_ be beneficiary for all clients.
Maybe we should have optional validations on the server?

bulk/repo-access-api.adoc
```diff
@@ -1,155 +1,335 @@
...
 
-Response:
+[NOTE]
+====
+link:https://www.rfc-editor.org/rfc/rfc9110.html#name-delete[RFC 9110: HTTP Semantics] states about the request body of a DELETE method:
+
+> An origin server SHOULD NOT rely on private agreements to receive content, since participants in HTTP communication are often unaware of intermediaries along the request chain.
 
-[source, json]
+Thus, this example sends the node ids via URL query.
```
enikao, line 263: Jos:: Weird to put them in the URL, instead of the body.

bulk/repo-access-api.adoc
```diff
@@ -1,155 +1,335 @@
...
+delete::
+* If node is a Partition: #TODO#
+* Otherwise: Assume the to-be-deleted node `D`,  has parent `P`.
++
+<<store, store call>> that sends `P` with all its features, but omits `D` from the children/annotations.
 
-. How do we represent the actual operations? Which transport protocol do we use?
+move:: Assume we want to move node `N` from its current parent `S` to its new parent `T`.
++
+<<store, store call>> that sends `T` with all its features, including `N` in the children.
```
enikao, line 332: Jos:: According to the explanation before `S` should also be sent.

enikao (commented):
Moved Jos' remarks from file contents to comments

bulk/repo-access-api.adoc
```diff
@@ -78,8 +90,7 @@ Creates new partitions in the repository.{fn-org216}
 Each sent node is its own partition.
 Thus, we cannot send the contents (i.e. (indirect) annotations/containments) of a partition; We can send them in a later <<store>> call.
 We also MUST NOT mention any annotation/containment node ids in the partition nodes, as they cannot be part of the same request, and we don't allow moving nodes in this operation.
-
-We MAY send properties and references #TODO correct?#{fn-org225}
+We MAY send properties and references{fn-org225}
 
 Each partition node id MUST NOT exist in the repository, and the sending client MUST use node ids allocated to it via <<ids>>.
```
joswarmer, line 103: This would be the only case where the node ids MUST be allocated through <<ids>>.  As with the other node ids, I think this should be optional.

enikao:
I closed all old discussions, as they are probably outdated

enikao:
As of last Friday, we're mostly done with this spec. Missing some editorial refinements and defining _message kinds_. We'll add the latter over time when implementing the first examples.

Issue #171, opened 2023-09-25
Create roadmap page


enikao:
It would be useful for public consumption to have a roadmap page, so they have an idea what's there and what's coming.

I don't think we can provide a timeline, but at least list the things we think about.

enikao:
Implemented, see https://lionweb-io.github.io/specification/roadmap/roadmap.html

Issue #28, opened 2022-10-28
Repo API: Change-based event notifications


enikao:
Aka delta protocol

# Use cases
* Editor sends fine-grained updates to repo
* Editor receives fine-grained updates from repo
  * Might want to inform user about "foreign" changes (e.g. highlight)
* Trigger processor recalculation (e.g. typesystem, checker)
* Trigger processor recalculation + update (e.g. bi-directional link processor: a processor that simulates bi-directional links by automatically updating two inverse "regular" references)

# Scenarios:
* Dumb repo: No delta support at all
* Unsynchronized repo: Whatever written last wins, conflicts blow up
  * Useful for single-user, or in combination with pessimistic locking
* Conflict-detecting repo (optimistic locking): Repo detects conflicting writes, but doesn't resolve them on its own. Might give client additional info to decide.
* Conflict-avoiding repo: Automatically resolves conflicts
  * Refuses conflicting change
  * OT-based
  * CRDT-based

# Requirements / wishes
* Independent of concrete conflict resolution approach
  * BUT: according to Sascha, only works if every client uses the same OT library

# Aspects
* Connected to locking
* Connected to versioning
* Connected to live collaboration
* Do clients send the same messages as the repo?
* Do clients update their own model, or send update request to server and only recieve update?

# Additional info
* [draft-toomim-httpbis-braid-http-02](https://datatracker.ietf.org/doc/html/draft-toomim-httpbis-braid-http) describes a generic protocol with similar properties

# Delta messages in different systems

## Comparison

  | MPS | | | EMF | Modelix | | | LIonCore-MPS | Automerge | Tree CRDT 1 | Tree CRDT 2
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
Operation | OpenAPI | SModel | VCS diff |   | server API | model operations | model area |   |   |   |  
set   reference | SReferenceChangeEvent | SModelReferenceEvent | SetReferenceChange |   | SetReferenceOpData | SetReferenceOp | ReferenceChangeEvent | ChangeLinkDelta | PutMap |   | rename
set   property | SPropertyChangeEvent | SModelPropertyEvent | SetPropertyChange | SET | SetPropertyOpData | SetPropertyOp | PropertyChangeEvent | ChangePropertyDelta | PutMap |   | rename
[...]

Issue #27, opened 2022-10-28
Repo API: Access control


enikao:
Access control on the repository

Possible modes:
* No control: everything read/write for everybody
* Whole repo: different roles might read, write, or read/write the whole repository
* Fine-grained: different roles might read, write, or read/write parts of the repository

Questions:
* How would merging work if user does not have write access to changed part of repo?
* Level of fine-grained access (#29)
* How to ensure we don't leak information through indirect channels (e.g. typesystem check)?

PR #174, opened 2023-09-25
First version of roadmap document

enikao:
implements #171

### DIFF
--- a/roadmap/roadmap.adoc
+++ b/roadmap/roadmap.adoc
@@ -1,0 +1,158 @@
+include::../shared/issue-footnotes.adoc[]
+
+:m3: ../metametamodel/metametamodel
+:refarch: ../reference-architecture/reference-architecture
+:serialization: ../serialization/serialization
+
+= LionWeb Roadmap, Releases, and History
+:toc:
+:toclevels: 4
+// :sectnums:
+// :sectnumlevels: 4
+
+== Roadmap
+As a volunteer-run Open Source effort, we cannot commit to a timeline.
+We're actively looking for corporate sponsors, with some prospects in sight.
+
+=== In progress
+
+[#bulk-protocol]
+==== Specifying bulk protocol
+We've decided{fn-org148} to specify protocols only by their conceptual features.
+The first such protocol will be the https://github.com/LionWeb-org/organization/blob/d8a54813c6bd224ef84e53a26ac7d60a45a866b5/bulk/repo-access-api.adoc[Bulk Repository access API].
+It is used to store and retrieve nodes in batches at the moment of invocation.{fn-org25}
+It is intended for CRUD operations on (larger) sets of nodes.
+Additionally, this API allows to list the available <<{m3}.adoc#partition, Partitions>>, get Languages known to the repository, and request unused <<{m3}.adoc#identifiers, node ids>>.
+
+=== Planned
+We highly encourage any participation in LionWeb.
+So if you'd like to work on some of these features, or even only have an urgent need for them: let us know!
+This will influence the feature's priority.
+
+The list below is in no particular order.
+
+[#delta-protocol]
+==== Delta protocol
+As described in our <<{refarch}.adoc#delta, reference architecture>>, via delta protocol a client receives an initial snapshot similar to the bulk case.
+Afterwards, the client continues to receive changes (aka deltas) from the repo; it also writes back changes continuously.{fn-org28}
+
+Open questions for this protocol include
+
+* Which messages do we support?
+For example, is _move a child from position 2 to position 5 in the same list_ the same as _move a child from parent A to parent B_ the same message kind?
+
+* Is the protocol the same for client -> repository updates vs. repository -> client updates?
+That's especially relevant for (web-based) editors.
+
+* How strong do we couple the protocol to an implementation?
+Both https://en.wikipedia.org/wiki/Operational_transformation[operational transformation] and https://en.wikipedia.org/wiki/Conflict-free_replicated_data_type[CRDTs] can be used to implement consistent clients.
+For both of them, several implementations are available.
+Additionally, we might support simpler clients that assume no conflicts or missing updates at all.
+
+==== Standard derived models
+We intend to support several common language (workbench) functionalities via <<{refarch}.adoc#derivedModel, derived models>>.
+For each functionality, LionWeb would define a _base language_.
+Any implementation might use these base languages directly, or define its own extension languages.
+
+As an example, LionWeb's _validation base language_ might define concepts for `Error`, `Warning`, and `Info`.
+My validator implementation would create instances of `Error` and `Info`, but also include a custom _compiler-based validation language_ that defines `GeneratorWarning` and `CompilerWarning` as specializations of `Warning`.
+A client that only knows about LionWeb's _validation base language_ can process both of the specialized warnings in a generic fashion.
+Another, compiler-aware client might display ``GeneratorWarning``s differently than ``CompilerWarning``s.
+
+_Base language_ candidates are typically common language functionalities that might be expensive to calculate.
+This includes:
+
+===== Validation
+Checking a model for consistency and other errors is one of the most common operations on models.
+It might also be expensive: if we wanted to assure globally unique names, we had to check all named things in our model on every name change.
+The basic structure of validation is very similar in most systems: any issue has a severity level, refers to a node, might relate to a feature, and provides an error message.
+
+===== Scoping
+Calculating the scope of visible targets at any given place is necessary for both parser-based clients (for name binding / linking) and projectional clients / IDEs (for auto-completion).
+Similar to validation, the baseic structure of scoping is very similar in most systems; this _base language_ would condense the common concepts.
+
+===== Typesystem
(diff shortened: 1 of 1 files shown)


roadmap/roadmap.adoc
```diff
@@ -0,0 +1,158 @@
...
+:toclevels: 4
+// :sectnums:
+// :sectnumlevels: 4
+
+== Roadmap
+As a volunteer-run Open Source effort, we cannot commit to a timeline.
+We're actively looking for corporate sponsors, with some prospects in sight.
+
+=== In progress
+
```
dslmeinte, line 18: Maybe something along the lines of finalizing the `2023.1` release, for completeness' sake?

roadmap/roadmap.adoc
```diff
@@ -0,0 +1,158 @@
...
+[#bulk-protocol]
+==== Specifying bulk protocol
+We've decided{fn-org148} to specify protocols only by their conceptual features.
+The first such protocol will be the https://github.com/LionWeb-org/organization/blob/d8a54813c6bd224ef84e53a26ac7d60a45a866b5/bulk/repo-access-api.adoc[Bulk Repository access API].
+It is used to store and retrieve nodes in batches at the moment of invocation.{fn-org25}
+It is intended for CRUD operations on (larger) sets of nodes.
+Additionally, this API allows to list the available <<{m3}.adoc#partition, Partitions>>, get Languages known to the repository, and request unused <<{m3}.adoc#identifiers, node ids>>.
+
+=== Planned
```
dslmeinte, line 27: "Planned" might be too committing... "Envisioned"?

roadmap/roadmap.adoc
```diff
@@ -0,0 +1,158 @@
...
+* Which messages do we support?
+For example, is _move a child from position 2 to position 5 in the same list_ the same as _move a child from parent A to parent B_ the same message kind?
+
+* Is the protocol the same for client -> repository updates vs. repository -> client updates?
+That's especially relevant for (web-based) editors.
+
+* How strong do we couple the protocol to an implementation?
+Both https://en.wikipedia.org/wiki/Operational_transformation[operational transformation] and https://en.wikipedia.org/wiki/Conflict-free_replicated_data_type[CRDTs] can be used to implement consistent clients.
+For both of them, several implementations are available.
+Additionally, we might support simpler clients that assume no conflicts or missing updates at all.
```
dslmeinte, line 50: And: Can we find a sensible common denominator to both implementation (OT, CRDT)?

roadmap/roadmap.adoc
```diff
@@ -0,0 +1,158 @@
...
+===== Typesystem
+We need to calculate the type of a node in several contexts, including validation (e.g. "do the types of these assigned nodes match?") and scoping (e.g. "which members does the thing on the left side of the dot have?").
+
+===== Commenting / Review notes
+For any kind of collaboration, it's very helpful for reviewers to attach their remarks directly to their context -- think of the blue speech bubbles in the right margin of a Word document.
+
+
+==== Access control
+Different users might have different level of access to the whole repository, or parts of it.
+Some users might have read-only access (e.g. via public web), whereas others can write to the repository.{fn-org27}
```
dslmeinte, line 83: Might be able to leverage existing systems for this cross-cutting yet orthogonal concern?

roadmap/roadmap.adoc
```diff
@@ -0,0 +1,158 @@
...
+
+==== Versioning / Live collaboration
+A repository might support version control and/or live collaboration of users (both features are technically related).
+The repository might have a version control backend like git, or support live collaboration like Google Docs.{fn-org26}
```
dslmeinte, line 87: Live collaboration would typically use the delta protocol. How does versioning relate to live collaboration?

roadmap/roadmap.adoc
```diff
@@ -0,0 +1,158 @@
...
+Support a common way how to handle language migrations, e.g. by specific messages like "I'm about to migrate this partition" or "You must not use language X in a version older than abc".{fn-org72}
+
+==== Store invalid text typed at arbitrary place
+Given a valid input with invalid text at few places, both tolerant parsers and projectional editors can provide a "mostly correct" AST representation.
+Internally, they have some way of keeping the invalid text around.
+We could support a standard way to store such invalid text in the repository.{fn-org62}
+
+==== Define a standard way to represent model diffs
+Calculating the difference between two sets of nodes is common, complex, and potentially expensive task.
+It might be useful to have a standard format to exchange the difference, and the <<delta-protocol>> could provide a solid base.{fn-org40}
```
dslmeinte, line 114: We might want to look at [EMF Compare](https://eclipse.dev/emf/compare/) for this. We could reimplement this, or re-use EMF Compare by exporting to EMF(/Ecore), running the comparison, and importing the result from EMF(/Ecore) to LionWeb again.

dslmeinte (approved):

roadmap/roadmap.adoc
```diff
@@ -0,0 +1,158 @@
...
+:serialization: ../serialization/serialization
+
+= LionWeb Roadmap, Releases, and History
+:toc:
+:toclevels: 4
+// :sectnums:
+// :sectnumlevels: 4
+
+== Roadmap
+As a volunteer-run Open Source effort, we cannot commit to a timeline.
```
ftomassetti, line 14: I would move that to the end, as the most interesting thing for readers would be the features we are working on, while this is to me an (important) foot-note

roadmap/roadmap.adoc
```diff
@@ -0,0 +1,158 @@
...
+We're actively looking for corporate sponsors, with some prospects in sight.
+
+=== In progress
+
+[#bulk-protocol]
+==== Specifying bulk protocol
+We've decided{fn-org148} to specify protocols only by their conceptual features.
```
ftomassetti, line 21: Perhaps we could add a sentence to explain what it will enables us and users to build. Integration with editors that do not have real-time collaboration?

roadmap/roadmap.adoc
```diff
@@ -0,0 +1,158 @@
...
+This will influence the feature's priority.
+
+The list below is in no particular order.
+
+[#delta-protocol]
+==== Delta protocol
+As described in our <<{refarch}.adoc#delta, reference architecture>>, via delta protocol a client receives an initial snapshot similar to the bulk case.
+Afterwards, the client continues to receive changes (aka deltas) from the repo; it also writes back changes continuously.{fn-org28}
+
+Open questions for this protocol include
```
ftomassetti, line 39: I am not sure if we need to include the open-questions in the roadmap

roadmap/roadmap.adoc
enikao: Hopefully! But we should have defined behavior if e.g. I'm asking for node id 17 with all descendants, but only have read access for 3 out of its 5 children.

roadmap/roadmap.adoc
enikao: AFAIK, all existing live collaboration implementations need a history of all edit actions to actually work. That's what I mean with "technically related".

roadmap/roadmap.adoc
dslmeinte: True. EMF Compare might have some interesting insights that could inform the precise protocol as well, though.

roadmap/roadmap.adoc
dslmeinte: I'm thinking we might introduce a generic `InaccesibleNode` concept in the `builtins` for that - slightly similar to the existing `Node` concept. That way the protocol can just return instances of that type instead of erroring out.

enikao: I mentioned only one example of lots of cases we should have defined behavior for. That's why there's #27 as a future item.

enikao:
In general, I'm not trying to add anything _new_ to the discussion here. This document is meant for the general public to have an idea what we're thinking about.

roadmap/roadmap.adoc
enikao: added

roadmap/roadmap.adoc
enikao: In line with Federico's comment https://github.com/LionWeb-org/organization/pull/174#discussion_r1336097035, I'm moving the details to #28

roadmap/roadmap.adoc
enikao: I agree this is not that important. But without this sentence the next one is hard to understand, and that one should be seen by everybody.

roadmap/roadmap.adoc
```diff
@@ -0,0 +1,158 @@
...
+The basic structure of validation is very similar in most systems: any issue has a severity level, refers to a node, might relate to a feature, and provides an error message.
+
+===== Scoping
+Calculating the scope of visible targets at any given place is necessary for both parser-based clients (for name binding / linking) and projectional clients / IDEs (for auto-completion).
+Similar to validation, the baseic structure of scoping is very similar in most systems; this _base language_ would condense the common concepts.
+
+===== Typesystem
+We need to calculate the type of a node in several contexts, including validation (e.g. "do the types of these assigned nodes match?") and scoping (e.g. "which members does the thing on the left side of the dot have?").
+
+===== Commenting / Review notes
```
enikao, line 77: I'm moving this to a separate section, as reviews _cannot_ be derived models -- we cannot reconstruct them from the original models.

Issue #104, opened 2023-03-31
Interfaces vs. (abstract) Concepts vs. Multiple Inheritance


enikao:
# Involved elements
In LIonCore M3, we might support both `Interface` and `Concept` as central elements.
`Concept` might be `abstract`.
`Interface` might support all kinds of *members*, or only a subset.
`Concept` might extend at most one other concept (single inheritance), or might extend more than one concept (multiple inheritance).

# Kinds of members:
* Property
* Containment
* Reference
* Derived feature (#6)

*Property*, *Containment*, and *Reference* are *features*.

# Alternatives
* Option A: Single inheritance (abstract) concepts, interfaces contain only derived features
* Option B: Single inheritance (non-abstract) concepts, interfaces contain anything
   i.e. interface is alias for abstract concept
* Option C: Multiple inheritance (abstract) concepts, no interfaces
* Option D: Multiple inheritance (abstract) concepts, no interfaces, but can extend only one non-abstract
  i.e. abstract concept is alias for interface
* Option E: Multiple inheritance (abstract) concepts, interfaces can contain anything

# Considerations

## Design aspects
* Interfaces must be sufficiently different from Abstract Concepts to warrant both their existence

## Semantic considerations
* Interfaces should represent “what something can do”, ie. a contract
* abstract concepts should represent “what something is”, i.e. a data holder

## Alignment with programming languages
Having an obvious mapping between LIonCore and popular programming languages is a worthy goal.
It's unclear how far we should go to reach that goal.

## Compatibility with single-inheritance language, e.g. Java, Kotlin
We want to convert between e.g. Java <--> LIonCore

To support multiple inheritance, we'd need to map concepts to interfaces, and generate the right implementations.
The generated code can get complex, and thus hard to debug.

### Option B
If we had *Interface* and (abstract) *Concept* with single inheritance in LIonCore, the mapping is straight-forward:
* Java `interface MyIface` <--> LIonCore `interface MyIface`
* Java `(abstract) class MyThing` <--> LIonCore `(abstract) concept MyThing`
* getter `public Collection<Name> getAliases()` <--> derived feature `aliases: [0..n] Name`
* getter `public int getAge()` + setter `public void setAge(int age)` + field (if inside a class) `private int age` <--> feature `age: Integer`

Advantage: Simple mapping, might be implemented dynamically.

### Option C
If we had (abstract) *Concept* with multiple inheritance and **no** *Interface*, mapping gets more involved, but still doable:
[...]

PR #197, opened 2023-10-16
rename interface, core languages with underscore

enikao:


dslmeinte (changes requested):
`misc/conceptinterface.adoc` should also be renamed, both the file name as well as the contents.

dslmeinte:
> Rename or delete? The contents are kind of outdated.

First one can be deleted, I suppose. Second one just needs renaming of contents.

enikao:
I changed my mind and kept both files, they are still relevant to the postponed discussion in #104.

Issue #172, opened 2023-09-25
Use term `release` instead of `version` for LionWeb?


enikao:
I'd like to disambiguate _version_ (as in version control) and _release_ (as in labeled, stable state of LionWeb).

We currently use the term `version` at:

Serialization:
1. SerializationChunk, field `serializationFormatVersion`
2. UsedLanguage, field `version`
3. MetaPointer, field `version`

Metamodel:
1. Language, field `version`

Of the above, I'd change `serializationFormatVersion` to `serializationFormatRelease`.

The other occasions refer to the user-controlled language version. I'm not sure whether we want to change them.

dslmeinte:
I like `serializationFormatRelease` because we're likely to have very few versions of that, and each of these probably doesn't need to go through a development/pre-lease lifecycle that benefits from being explicit in the version ID.

For the other things, it's much more likely there's a development/pre-lease lifecycle that benefits from being explicit, so I'd be hesitant to flip that over to `<xxx>Release` just yet.

PR #210, opened 2023-11-20
Persist version 2023.1 contents explicitly

enikao:


2023.1/index.adoc
```diff
@@ -0,0 +1,37 @@
...
+<<{m3}.adoc#Integer, Integer>>,
+and <<{m3}.adoc#JSON, JSON>>.
+
+== Contents
+
+* Serialization format{emquad}<<{serialization}.adoc, html>>{emquad}link:metametamodel/metametamodel.pdf[pdf]
+
+* Meta-metamodel{emquad}<<{m3}.adoc, html>>{emquad}link:serialization/serialization.pdf[pdf]
+
+* #TODO: Also reference architecture?#
```
enikao, line 36: Should we include the reference architecture? It might change in the future, and we refer to it.

2023.1/serialization/serialization.adoc
```diff
@@ -0,0 +1,519 @@
...
+== Versions
+
+NOTE: The term "version" is a bit ambiguous within LionWeb.{fn-org172}
+This section lists all versions as they appear in <<SerializationChunk.serializationFormatVersion>> and <<{m3}.adoc#Language.version, `Language.version`>> for languages `LionWeb_M3` and `LionWeb_builtins`.
+
+=== 2023.1
+Technical name: `2023.1`
+
+Initial version.
+Refer to <<{roadmap}.adoc#release-2023.1, roadmap>> for details.
```
enikao, line 432: Wrong reference

2023.1/serialization/serialization.adoc
```diff
@@ -0,0 +1,519 @@
...
+[[json-schema]]
+== JSON Schema for serialization
+[source,json]
+----
+include::serialization.schema.json[]
+----
+
+[[possible-values]]
+== Possible values for `properties`, `containments`, and `references`
+#TODO Shall this stay in?#
```
enikao, line 443: Do we want to include this? I guess nobody ever read it.

enikao:
Shall we use term "version" or "release"?

2023.1/index.adoc
dslmeinte: Referring to it would be good, IMHO. I don't think we need to version the reference architecture (explicitly).

dslmeinte:
> Shall we use term "version" or "release"?

Or "release[d] version"? But I think I'd prefer "release". (We could always vote on it via Slack.)

2023.1/serialization/serialization.adoc
dslmeinte: Doesn't matter: if it's part of the specification (or a clarification/elucidation of it), then it can stay – even just in case someone later says "[...]but you didn't properly specify this!"

dslmeinte (approved):

2023.1/serialization/serialization.adoc
enikao: "Nobody ever read" also increases the chance it's incorrect ...

2023.1/index.adoc
enikao: Then we'd need to care for it in the case of future changes. And nobody will remember that.

enikao:
> > Shall we use term "version" or "release"?
> 
> Or "release[d] version"? But I think I'd prefer "release". (We could always vote on it via Slack.)

We actually had an issue on this: https://github.com/LionWeb-io/specification/issues/172

-> Don't use "release"

2023.1/serialization/serialization.adoc
dslmeinte: That's absolutely true, unfortunately... Could we somehow set it apart for the next release, and not include it in the current?

2023.1/index.adoc
dslmeinte: True: I guess it's best to remove it, then.

dslmeinte:
> > > Shall we use term "version" or "release"?
> > 
> > 
> > Or "release[d] version"? But I think I'd prefer "release". (We could always vote on it via Slack.)
> 
> We actually had an issue on this: #172
> 
> -> Don't use "release"

OK, so it _has_ to be "version" then, right?

2023.1/serialization/serialization.adoc
enikao: So we link to the "living" version now, but earmark it for the next release?

dslmeinte (approved):

2023.1/serialization/serialization.adoc
enikao: > So we link to the "living" version now, but earmark it for the next release?

My bad, I mixed up the two discussion threads.

I now removed the "possible values" section from the 2023.1 document (still in the living version).

PR #231, opened 2024-01-28
Slides from LangDev 2023 and Strumenta Panel 2024

enikao:


dslmeinte (approved):
Maybe export the `.pptx` as a PDF?

enikao:
> Maybe export the `.pptx` as a PDF?

They are in, with slightly different name (to be recognized out of context).

PR #232, opened 2024-01-29
Fix the builtins and M3 jsons 

joswarmer:
In the json for M2 and builtins:

- builtins was missing as used language => added
- language-version was "1", changed to "2023.1"

### DIFF
--- a/metametamodel/builtins.json
+++ b/metametamodel/builtins.json
@@ -4,6 +4,10 @@
     {
       "key": "LionCore-M3",
       "version": "2023.1"
+    },
+    {
+      "key": "LionCore-builtins",
+      "version": "2023.1"
     }
   ],
   "nodes": [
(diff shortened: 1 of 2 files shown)


metametamodel/builtins.json
```diff
@@ -4,6 +4,10 @@
     {
       "key": "LionCore-M3",
       "version": "2023.1"
+    },
+    {
+      "key": "LionCore-builtins",
```
dslmeinte, line 9: Wasn't the dependency on this considered to be implicit?

joswarmer: We discussed this several times and finally made this decision:
image
from the serialization document.

dslmeinte: Ah, right. Apologies for not looking this up in the spec. myself!

dslmeinte (approved):

enikao:
Great find!

enikao:
> (Why can I edit other people's comments?!)

Because you're repo owner

dslmeinte:
> > (Why can I edit other people's comments?!)
> 
> Because you're repo owner

Me, repo owner?! Who goofed that up?!?! 😃

PR #235, opened 2024-02-04
Ideas on derivation

enikao:
based on #167 

Very unfinished, just scribbling down my thoughts.
This draft PR is meant as a place to collect comments.


derived/derivation.adoc
```diff
@@ -90,6 +90,11 @@ IDerived "0..*" -> "1..*"  Node: base
 Each derived node references one or more base nodes.
 Any (base) node can have none or more derived nodes.
 
+[horizontal,labelwidth=12]
+QUESTION:: Should a derived node also be able to refer to features inside a node? Or do we leave that for specific classes implementing `IDerived`?
```
enikao, line 94: In the example below, the `PropertyViolation` points to a specific feature.

Assume the following concepts:
```
concept Person [id=P, key=pers]
  property firstName: string [id=PF, key=firstN]
  property lastName: string [id=PL, key=lastN]

concept EmptyNameViolation implements IDerived, IViolation [id=ENV, key=emptyNameV]
  property message: string [id=ENVM, key=msg]
  reference property: Property [id=ENVP, key=prop]
```

Original model contains instance:
```
Person A [id=2]
  firstName = "John"
  lastName = ""
```

Validation derived model contains instance:
```
EmptyNameViolation B [id=5]
  message = "lastName is empty!"
  base = --> A[id=2]
  property = --> lastName[id=PL]
```

derived/derivation.adoc
```diff
@@ -90,6 +90,11 @@ IDerived "0..*" -> "1..*"  Node: base
 Each derived node references one or more base nodes.
 Any (base) node can have none or more derived nodes.
 
+[horizontal,labelwidth=12]
+QUESTION:: Should a derived node also be able to refer to features inside a node? Or do we leave that for specific classes implementing `IDerived`? 
+QUESTION:: The assumption is that derived nodes can have children that are not derived nodes.
```
enikao, line 95: Derived nodes can have any kind of features, including children that might or might not implement `IDerived`.

derived/derivation.adoc
```diff
@@ -134,6 +139,9 @@ We need to omit the parameter if we don't want to limit the depth.
 ==== Result
 {chunk} containing all derived nodes according to `nodeIds`, `languages`, and `depthLimit` parameters.
 
+[horizontal,labelwidth=12]
+QUESTION:: The `depth:Limit` describes the depth of the retrieved nodes (derived nodes may be full trees), the next line suggests that `depthLimit` applies to the node id's sent. Needs clarification.
```
enikao, line 143: The idea is:
1. We select the same original nodes as with Bulk API retrieve
2. For all these nodes, we find their derivations (if available)
3. For all derivations, we return the derivations and all descendants thereof with infinite depth.
   The assumption is that derived nodes only contain other nodes that are part of the derived information.

derived/derivation.adoc
```diff
@@ -143,6 +151,8 @@ Does NOT include the definition of <<{serialization}.adoc#UsedLanguage, UsedLang
 
 
 == Possible derivation backends
+[horizontal,labelwidth=12]
+QUESTION:: The repository that is referred to is this always the repository where the opriginal model is stored, or could oit be a different (i.e. derivation processor specific)  repository?
```
enikao, line 155: Formally the derivation backend implementation is independent of the repository implementation, i.e. they don't need to be the same "service". (In practice in makes a lot of sense to couple them, though.)

Derived models are part of the same repository as the original model in the sense that they share one id-space, and can refer to each other.

derived/derivation.adoc
```diff
@@ -270,6 +280,9 @@ package M3 {
 }
 ----
 
+[horizontal,labelwidth=12]
+QUESTION:: Should the `PropertyValidation`  also point to the Property instance in the M2 model being validated, and not just to the node that contains the property?
```
enikao, line 284: See remark above.

ftomassetti:
I am not sure if there is a better place to write these comments, but reading the current document on derived models I had these questions/comments:

* Can other sources be involved in creating a derived node (e.g., external data)? So should we say that a derived node is produced by examining existing nodes (both "original nodes" and other derived nodes) plus, potentially, external sources?
* Can we have a derived partition or just other nodes?
* What about changing the term base node in source node? 
* Should derived nodes have some sort of marker to figure out which derivation produced them? For example, an annotation
* Should derivations have a name or some identifier?
* Why do we need the IDerived interface? Can't I derive any node I want? For example, I may want to use derive nodes to de-sugar code, without having to mark all of my expressions or statements as implementing IDerived

enikao:
> Can other sources be involved in creating a derived node (e.g., external data)? So should we say that a derived node is produced by examining existing nodes (both "original nodes" and other derived nodes) plus, potentially, external sources?

Yes, as long as the processor creating the derived nodes takes care of interacting with the external source.

> Can we have a derived partition or just other nodes?

I'm not sure I understand. We can have a derivation of any node, not only partitions.

> What about changing the term base node in source node?

I don't really mind; I think we used this term ` base` already at other places for this idea.

> Should derived nodes have some sort of marker to figure out which derivation produced them? For example, an annotation

Derivations are defined by their _ava_, i.e. a language. That might be enough of a marker -- or not.

> Should derivations have a name or some identifier?

Derivations in sense of derived nodes are marked by implementing `IDerived`.
Derivations themselves are defined by their _ava_, i.e. a language.
Do we need more identification?

> Why do we need the IDerived interface? Can't I derive any node I want? For example, I may want to use derive nodes to de-sugar code, without having to mark all of my expressions or statements as implementing IDerived

`IDerived` connects the derived node to its base. Without that, how do we know from what we derived it?
We can have an `EvaluationResult implements IDerived` that refers to the evaluated expression.
The `EvaluationResult` can contain arbitrary complex sub-nodes to e.g. describe the execution time of each sub-expression. These sub-nodes don't need to implement `IDerived` (even though it might make sense, because we can connect them uniquely to a base node).

derived/processors.pptx
dslmeinte, line 1: @joswarmer Can you also commit a PDF version of this?

joswarmer: There is a PNG version next to the PowerPoint

dslmeinte: Sorry, figured the PowerPoint had more than the one figure :)

derived/processors.adoc
```diff
@@ -0,0 +1,133 @@
...
+- In a derived model all nodes are created (automatically) by a processor. 
+  It contains derived nodes and calculated nodes.
+
++
+However, a derived model MAY contain nodes that do not implement `IDerived`.
+Example: A derived richtext description contains `Word` concepts.
+`Word` is reused in many contexts, and does not implement `IDerived`.
+In this case `Word` is a calculated node, but *not* a derived node.
+
+NOTE: A repository always contains exactly one original model and any number of derived models.
```
dslmeinte, line 48: Only one original model? Not any number of models?

derived/processors.adoc
```diff
@@ -0,0 +1,133 @@
...
+
+NOTE: As defined here, nodes that are calculated by a processor can never be children or annotations of an original node.
+
+base model::
+A model that has _derived_ models.
+A base model can be an _original_ or a _derived_ model.
+
+processor::
+A _processor_ is a component that provides/calculates derived nodes for nodes in a base model.
```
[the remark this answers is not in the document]
joswarmer: I think a processor should have a set of partitions for which it derived nodes. Which of the nodes in the partitions will havederived nodes is up to the processor.

derived/processors.adoc
joswarmer: It contains any number of original partitions.

PR #310, opened 2024-10-20
article: Anatomy of a LionWeb model

enikao:


articles/model-anatomy.adoc
```diff
@@ -0,0 +1,644 @@
...
+annotation review as "33:ReviewComment"
+
+steering *- review: <<annotations>>
+
+engine -[hidden]> steering
+----
+
+In summary, a node's parent must always be symmetric -- either to the parent's children, or the parent's annotation instances.
+
+If we deleted the parent node, we implicitly delete all its children and annotation instances, each of their children and annotation instances, and so on.
```
dslmeinte, line 153: "[...]we implicitly delete**d** all [...]"

articles/model-anatomy.adoc
```diff
@@ -0,0 +1,644 @@
...
+
+The following subsections describe the kinds of data types supported by LionWeb.
+
+==== Primitive types
+LionWeb ships with built-in types for `String`, `Integer`, and `Boolean`, as they are needed in almost all systems.
+String properties can have any string value, including empty.
+Integer properties support positive and negative decimal integers of any length.
+Boolean properties can only have the values `true` or `false`.
+A language might define its own additional _primitive types_.
+In any case, we store all of them as string on low-level serialization format (i.e. JSON).
```
dslmeinte, line 198: "_on_ low-level serialization format" doesn't sound entirely right to me

articles/model-anatomy.adoc
```diff
@@ -0,0 +1,644 @@
...
+json "TextRange [key=tr]" as tr {
+  "start [key=st]": "TextPosition",
+  "end [key=en]": "TextPosition",
+  "label [key=la]": "String",
+  "valid [key=va]": "Boolean"
+}
+----
+
+We store the value of a structured data type as string, the same as all other primitive types.
+The content of that string is an escaped JSON structure.
```
dslmeinte, line 243: "is a JSON structure serialized as a JSON string" would be more accurate, I think.

articles/model-anatomy.adoc
```diff
@@ -0,0 +1,644 @@
...
+The same target can appear several times in a multiple reference.
+This can help, for example, to describe steps: `steps: Task[1..*] = {generate, saveAll, compile, saveAll}` refers to Task `saveAll` twice.
+
+LionWeb does not support bi-directional references.
+They are very hard to maintain, and also hard to process (i.e. if we want to traverse the whole model).
+
+==== ResolveInfo
+For references, we store the target node's id, and additionally a `resolveInfo`.
+ResolveInfo is a string that somehow describes the target node.
+Typical candidate would be the target's node `name` (if the target node has such a property).
```
dslmeinte, line 389: Or alternatively, some format of qualified name?

articles/model-anatomy.adoc
```diff
@@ -0,0 +1,644 @@
...
+But in parser-based systems, after parsing we know only some kind of identifier of a reference target.
+We might want to store this unresolved _abstract syntax tree_ (AST).
+A later linking step hopefully finds the actual target and fills in the proper target node id.
+
+Target node id, resolveInfo, or both must be set for each reference -- otherwise the reference is pointless.
+
+==== Unresolved references
+Assume a node `T` that happens to be the target of a reference `myRef` from node `A`.
+When we delete `T`, the reference stays untouched: `A` still contains a reference `myRef` with target node id `T` (and maybe some resolveInfo), but the target cannot be resolved.
+LionWeb allows references with unresolvable target.
```
dslmeinte, line 411: It's up to the consumer of the model to do something with/about that target, though.

articles/model-anatomy.adoc
```diff
@@ -0,0 +1,644 @@
...
+  class A
+  class T<<h>>
+  A .> T: myRef
+}
+----
+
+=== Differences between containment and reference
+
+A contained node shares the lifetime of its parent -- if the parent is deleted, the contained node gets deleted, too.
+The contained node makes no sense without its parent.
```
dslmeinte, line 445: Well, it could make sense (to some degree), but there's no "place to live" for the contained node.

articles/model-anatomy.adoc
```diff
@@ -0,0 +1,644 @@
...
+In contrast, the target of a reference is perfectly valid without the source.
+They can be deleted independently.
```
dslmeinte, line 447: , and don't need to be deleted when sources pointing at them are deleted.

articles/model-anatomy.adoc
```diff
@@ -0,0 +1,644 @@
...
+
+Containments form a model into tree shape.
```
dslmeinte, line 449: Sentence feels a bit weird. I liked the "establish" formulation used earlier better.

articles/model-anatomy.adoc
```diff
@@ -0,0 +1,644 @@
...
+With references, we turn that tree into a graph, as a node can be target of several references.
```
dslmeinte, line 450: "regard as a graph" instead of "turn that tree into a graph"?

articles/model-anatomy.adoc
```diff
@@ -0,0 +1,644 @@
...
+For references, the source node stores the target node's id and/or resolveInfo.
+
+LionWeb does not support unresolvable containments, but unresolvable references are allowed.
+
+[[annotations]]
+=== Annotation instances
+Annotation instances are mostly regular nodes, with two exceptions:
+Their classifier must be an annotation, it cannot be a concept.
+Also, the annotation instance must be mentioned in its parent's annotation instance list, it cannot be mentioned in the parent's containments.
+Both these criteria are indirect; thus, when only looking at a node (without examining its classifier or parent), we cannot distinguish annotation instances from other nodes.
```
dslmeinte, line 489: Except for by looking at the annotation classifier itself.

dslmeinte (changes requested):

articles/model-anatomy.adoc
```diff
@@ -0,0 +1,644 @@
...
+Rectangle *-- Coordinate: topLeft
+Rectangle *-- Coordinate: lowerRight
+----
+
+A node's classifier describes the valid containments.
+This description includes the containment’s _name_, _key_, whether it is _optional_, whether we allow _multiple_ nodes in the containment, and its _type_.
+Name, key and optional mean the same for containments and <<properties, properties>>.
+A singular containment allows only one contained node, e.g. a `Rectangle` can contain only one `topLeft` coordinate.
+In contrast, a `House` can contain multiple ``Room``s.
+For multiple containments, we keep track which child is contained at which position.
```
Ulyana-F1re, line 293: Why not to call it an ordered set?

articles/model-anatomy.adoc
```diff
@@ -0,0 +1,644 @@
...
+
+On serialization level, we store the contained node's id for a containment.
+A containment is invalid if the contained node's id cannot be resolved.
+
+[[references]]
+=== References
+
+A node can point to other nodes it has some relation to.
+For example, a `Project` might refer to its `mainResponsible: Employee`, or to all its `externalParties: BusinessPartner`.
+The node containing the reference is called _source_, the referred node _target_ of the reference.
```
Ulyana-F1re, line 337: the referred node **is called** the _target_ of the reference

articles/model-anatomy.adoc
```diff
@@ -0,0 +1,644 @@
...
+This can help, for example, to describe steps: `steps: Task[1..*] = {generate, saveAll, compile, saveAll}` refers to Task `saveAll` twice.
+
+LionWeb does not support bi-directional references.
+They are very hard to maintain, and also hard to process (i.e. if we want to traverse the whole model).
+
+==== ResolveInfo
+For references, we store the target node's id, and additionally a `resolveInfo`.
+ResolveInfo is a string that somehow describes the target node.
+Typical candidate would be the target's node `name` (if the target node has such a property).
+To support this typical default behavior, LionWeb has a built-in interface `INamed` with one property `name: String` that might be implemented by any classifier.
```
Ulyana-F1re, line 390: This sentence comes a bit out of the blue: there has been no introduction of interfaces and of built-in classifiers so far. Btw this built-in interface `INamed` is something that was quite hard to match in Rascal (while even the built-in `Node` has the corresponding type in Rascal).

articles/model-anatomy.adoc
```diff
@@ -0,0 +1,644 @@
...
+However, it's up to the application to select and set the appropriate resolveInfo.
+
+This serves two purposes:
+
+* In the space project example above, assume some app that displays the list of external parties for the project as `ESA: 23, JAXA: 42`.
+Now someone thinks there's no project left with ESA, and deletes node 87 (the ESA one).
```
Ulyana-F1re, line 396: I had troubles parsing 'space project example above'. Maybe 'in the example above that describes a project' (there is nothing space-specific there)?

articles/model-anatomy.adoc
```diff
@@ -0,0 +1,644 @@
...
+If the `externalParties [0]` reference only knew the target node id, the best possible display would be `!unknown target 87!, JAXA: 42` -- not very helpful to the user.
```
Ulyana-F1re, line 397: display **of the `externalParties` reference** would be
(otherwise it's strange to see `JAXA: 42` there)

articles/model-anatomy.adoc
```diff
@@ -0,0 +1,644 @@
...
+M2 -[hidden]- M1
+
+'Function <.. f1: <<classifier>>
+'Argument <.. a1: <<classifier>>
+'SourcePosition <.. an1: <<classifier>>
+'SourcePosition <. an2: <<classifier>>
+
+----
+
+An annotation that is applicable to all concepts says it annotates `Node` from LionWeb's built-in library.
```
Ulyana-F1re, line 578: What about a section for the built-in library? (it was mentioned but not introduced before)

articles/model-anatomy.adoc
```diff
@@ -0,0 +1,644 @@
...
+Annotations can _annotate_ other annotations; however, this approach introduces complexities and should be used with great caution.
+
+#TODO:# Do we allow unresolvable annotations?
+
+=== Differences between containment and annotation instance
+Both a contained node `C` and an annotated node `A` share the lifecycle of their parent node `P`.
```
Ulyana-F1re, line 584: Not am annotated node, but an annotation instance, right?

articles/model-anatomy.adoc
```diff
@@ -0,0 +1,644 @@
...
+So what's the difference?
+
+A containment must be described in the parent's node classifier.
+If the parent is of classifier `House` with containments `rooms`, and `doors`, we cannot add a containment `windows` to the parent node.
```
Ulyana-F1re, line 588: we - as users of this language? Because as a designer, I could add `windows`.

articles/model-anatomy.adoc
```diff
@@ -0,0 +1,644 @@
...
+The designer of classifier `House` made the choice that for their domain, a house should only care about rooms and doors, but not windows.
+The designer might not have anticipated our use case of their domain.
+But every user of that classifier relies on this structure of a house, so we must not mess with it.
+
+An annotation instance cannot be mentioned in the parent's node classifier at all.
+Instead, an annotation instance's classifier must be an annotation, which mentions the classifier it annotates.
+Thus, the designer of the annotation decides which targets their annotation is applicable to.
+So we can come up with an annotation `WindowAnnotation` that `annotates: House`, and attach instances to nodes of classifier `House`.
+All these nodes still adhere to the intents of ``House``'s designer, but also transport the additional information about windows we need for our use case.
```
Ulyana-F1re, line 597: Do orthogonal nodes adhere to the original intents, or rather do not mess up with them?

articles/model-anatomy.adoc
```diff
@@ -0,0 +1,644 @@
...
+Similarly, we could come up with an annotation `ReviewComment` to be used during code reviews.
+The reviewer can attach instances to any node, without interfering with the original intent of these nodes.
+
+Review comments demonstrate one drawback of annotations: we're violating the generally valued idea of _separating concerns_, as we mix domain concerns and process concerns in the same model.
+The alternative would be to keep both concerns in separate models, and have the review comments reference the node they apply to.
+Experience shows this approach is very hard to maintain:
+What happens if we deleted a house, and there's still a review comment referring to it?
+The review model might not be loaded during the deletion, so it now contains a "zombie" review comment.
+We might need to introduce "garbage collection" strategies to find such zombies and deal with them.
+For more context-sensitive annotations, moving the annotated node might introduce similar problems.
```
Ulyana-F1re, line 612: Does this paragraph explain why on the M2 nodes contain their annotations?

articles/model-anatomy.adoc
```diff
@@ -0,0 +1,644 @@
...
+
+Annotations should not be used as replacement for proper concept design.
+If our language dealt with the domain of houses, review comments are completely independent of that domain, should not be part of it, and are perfect candidates for annotations.
+However, if our language was modeling discussions, then reviews and comments are integral part of that domain, and should be represented by proper containments and (abstract) concepts in our language.
+In this case, using annotations only because they seem conveniently applicable at lots of places is a strong modeling smell.
```
Ulyana-F1re, line 617: What about the House example above: it looks like windows should not be in annotations of houses?

Ulyana-F1re (commented):
Good text with the important overview of the decisions (pragmatics) behind the design of LionWeb. I tried to look mostly at the flow of the description.

articles/model-anatomy.adoc
```diff
@@ -0,0 +1,644 @@
...
+Then the question transforms to: What is a node?
+
+[[nodes]]
+== Nodes
+A node is the fundamental building block in LionWeb.
+It consists of a unique identifier (**id**), a **classifier** to define the nature / type / class of the node, the node's **parent** node and lists of **properties**, **containments**, **references**, and **annotation instances**, respectively.
+We group a node's properties, containments, and references as the **features** of that node.
+
+Properties are simple values like `name: String`, `age: Integer` or `alive: Boolean`.
+Containments are other nodes that this node is composed of; examples would be `mainEntrance: Door[1]` or `corners: Coordinate[0..\*]`.
```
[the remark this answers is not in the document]
enikao: My idea was to not explain too much here, as it's the absolute summary of what a node is. The multiplicities should slightly hint at the differences.

articles/model-anatomy.adoc
```diff
@@ -0,0 +1,644 @@
...
+References are other nodes that this node relates to; examples would be `surface: Material[1]` or `stakeholders: Employee[1..*]`.
+Annotation instances are other nodes that represent orthogonal concerns of this node; examples include `ParsedFromLine` or `ReviewComment`.
+
+Let's look into each of these in more detail.
+
+[[node-id]]
+=== Node id
+The id of a node is a string of arbitrary length, consisting of characters `A`-`Z`, `a`-`z`, `0`-`9`, dash `-` and underscore `_`.
+It cannot be empty, and it must be unique for all nodes in the model.
```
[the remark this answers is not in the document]
enikao: I don't want to introduce the idea of repository here.

articles/model-anatomy.adoc
```diff
@@ -0,0 +1,644 @@
...
+=== Meta-pointer
+Before we get to the next part of a node, classifier, we need to discuss _meta-pointers_.
+Any time we refer from an instance model (M1) to its meta-model (M2 a.k.a. _Language_), we use meta-pointers.
+They are a triple of {_language key_, _language version_, _element key_}, for example `{language=MyLanguage, version=23-alpha, key=MyConcept}`.
+Both language key and element key use the same character set as <<node-id, node ids>>, whereas language version can be any non-empty string.
+
+We need this triple, because a language can be uniquely identified by its key and version, and the element key must be unique within its language.
+With language version we support _language evolution_, i.e. the fact that real-world languages tend to change over time.
+
+We don't prescribe anything about language version, as different implementations might use different ways to represent language evolution.
```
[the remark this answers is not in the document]
enikao: That's more precise, but harder to read -- we also don't go to that length for language key/version a few lines above.

articles/model-anatomy.adoc
```diff
@@ -0,0 +1,644 @@
...
+This means the node can only have features as described in the node's classifier.
+footnote:[At least by default; we might deviate from that constraint under special circumstances, like migrations.]
+
+[[parent]]
+=== Node parent
+The parent of node `A` is the node `B` that contains `A`.
+For example, node `10` of classifier `Car` might contain node `5` of classifier `PetrolEngine`, and node `11` of classifier `SteeringWheel`.
+Thus, the parent of `5` is `10`, and the parent of `11` is also `10`.
+
+[plantuml, carParent, svg]
```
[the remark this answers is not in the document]
enikao: I'd expect the Asciidoctor pipeline would pick that up, as it does for all the other spec docs.

articles/model-anatomy.adoc
enikao: maybe "_in_ low-level serialization format"? I try to avoid calling it directly JSON, as that might be confused with the JSON type we used to have.

articles/model-anatomy.adoc
enikao: I tried quite a bit to describe this accurately here, but didn't succeed. That's why this is the only place where I refer to the spec.

What about "is a JSON object as escaped string"?

articles/model-anatomy.adoc
enikao: Would it? Because then I encode some kind of scoping into ResolveInfo, and it wouldn't work any more if the target was moved around.

articles/model-anatomy.adoc
enikao: Isn't this clarified sufficiently by the parenthesis?

articles/model-anatomy.adoc
```diff
@@ -0,0 +1,644 @@
...
+The classifier can either be a _concrete concept_ or an _annotation_. It cannot be an _abstract concept_ or _interface_, as we cannot instantiate them.
+
+A node must adhere to its classifier.
+This means the node can only have features as described in the node's classifier.
+footnote:[At least by default; we might deviate from that constraint under special circumstances, like migrations.]
+
+[[parent]]
+=== Node parent
+The parent of node `A` is the node `B` that contains `A`.
+For example, node `10` of classifier `Car` might contain node `5` of classifier `PetrolEngine`, and node `11` of classifier `SteeringWheel`.
```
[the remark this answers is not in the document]
enikao: The idea is to implicitly distinguish between ids (numbers) and keys (short names), also to illustrate that ids are meaningless.
This doesn't work for you?

articles/model-anatomy.adoc
```diff
@@ -0,0 +1,644 @@
...
+We need this triple, because a language can be uniquely identified by its key and version, and the element key must be unique within its language.
+With language version we support _language evolution_, i.e. the fact that real-world languages tend to change over time.
+
+We don't prescribe anything about language version, as different implementations might use different ways to represent language evolution.
+We just assume that equal strings refer to the identical version.
+
+=== Node classifier
+A node's classifier is a <<meta-pointer>> that defines the node's nature, or type or class -- "what kind of thing is this node?"
+A classifier of a node is equivalent to the class of a Java object.
+The classifier can either be a _concrete concept_ or an _annotation_. It cannot be an _abstract concept_ or _interface_, as we cannot instantiate them.
```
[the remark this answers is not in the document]
enikao: I added a section on notation at the beginning.

articles/model-anatomy.adoc
```diff
@@ -0,0 +1,644 @@
...
+car *- pos: <<annotations>>
+
+annotation review as "33:ReviewComment"
+
+steering *- review: <<annotations>>
+
+engine -[hidden]> steering
+----
+
+In summary, a node's parent must always be symmetric -- either to the parent's children, or the parent's annotation instances.
```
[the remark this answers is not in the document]
enikao: I mean "to every parent, there must be a counterpart (either containment or annotation)"
Same in the other direction: "to every containment, there must be a parent" / "to every annotation, there must be a parent".

We cannot have a parent without containment/annotation, or containment without parent, or annotation without parent.

articles/model-anatomy.adoc
```diff
@@ -0,0 +1,644 @@
...
+class person as "3:Person" {
+  name: String = "Joe"
+  age: Integer = 23
+  alive: Boolean = true
+}
+----
+
+A node's classifier describes which properties can appear in the node.
+This description includes the property's _name_, _key_, whether it is _optional_, and its _type_.
+The name is for humans to understand the property; we use the key to uniquely identify it technically.
```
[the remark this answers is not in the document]
enikao: All of this is about the property: property name for humans, property key for computers.

articles/model-anatomy.adoc
```diff
@@ -0,0 +1,644 @@
...
+class Rectangle
+
+class Coordinate
+
+Rectangle *-- Coordinate: topLeft
+Rectangle *-- Coordinate: lowerRight
+----
+
+A node's classifier describes the valid containments.
+This description includes the containment’s _name_, _key_, whether it is _optional_, whether we allow _multiple_ nodes in the containment, and its _type_.
```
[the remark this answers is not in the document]
enikao: I struggle whether I should keep the meta-levels neatly separated or mash them together. The spec docs separate, in this article I tried the mashing. Not sure which one is easier to understand.

I might add a bit more description of M2 to the last section.

articles/model-anatomy.adoc
```diff
@@ -0,0 +1,644 @@
...
+Name, key and optional mean the same for containments and <<properties, properties>>.
+A singular containment allows only one contained node, e.g. a `Rectangle` can contain only one `topLeft` coordinate.
```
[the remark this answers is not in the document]
enikao: added "(i.e. `multiple = false`)"

articles/model-anatomy.adoc
enikao: does the link help?

articles/model-anatomy.adoc
enikao: technically, "contained" is equivalent to "annotated node" (as established above). Not sure which one is easier to understand here.

articles/model-anatomy.adoc
enikao: Good idea

articles/model-anatomy.adoc
enikao: Is this question answered in the next paragraph?

articles/model-anatomy.adoc
enikao: I don't get the question: All annotation instances are contained in their annotated node, both on M1 and M2.

articles/model-anatomy.adoc
enikao: good point, added some discussion.

articles/model-anatomy.adoc
dslmeinte: Then maybe "_in **the**_ low-level serialization format"?

articles/model-anatomy.adoc
dslmeinte: True, and indeeds suggests adding meaning to these things.

articles/model-anatomy.adoc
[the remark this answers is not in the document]
dslmeinte: OK, I'll assume that that'll work then

articles/model-anatomy.adoc
```diff
@@ -24,6 +24,13 @@ We deliberately don't use that term in our specifications, as it's so overloaded
 For our purposes, we can always replace "model" with "a bunch of nodes".
 Then the question transforms to: What is a node?
 
+[[notation]]
+== Notation used in this article and diagrams
```
Ulyana-F1re, line 28: I would put this section after the introduction of the node ids and their classifiers, so in between 2.3 `Node classifier` and 2.4 `Node parent`. There it is more logical for the reading flow, and just before it is being used.

articles/model-anatomy.adoc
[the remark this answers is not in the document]
Ulyana-F1re: I would rephrase then into `In summary, a node's parent must always be a counterpart -- either to the parent's children, or the parent's annotation instances`. 'Symmetric' lands differently for me: "symmetric = made up of exactly similar parts facing each other or around an axis". So it's not the node's parent that is symmetric.

articles/model-anatomy.adoc
```diff
@@ -0,0 +1,644 @@
...
+
+class Coordinate
+
+Rectangle *-- Coordinate: topLeft
+Rectangle *-- Coordinate: lowerRight
+----
+
+A node's classifier describes the valid containments.
+This description includes the containment’s _name_, _key_, whether it is _optional_, whether we allow _multiple_ nodes in the containment, and its _type_.
+Name, key and optional mean the same for containments and <<properties, properties>>.
```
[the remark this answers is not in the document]
Ulyana-F1re: I guess this sentence just doesn't parse well: 'optional' is without object. Font highlighting could help with this. 
"The name, key, and optional flag have the same meaning as described above for properties. Comparing to the properties though, a containment can be singular or multiple."

articles/model-anatomy.adoc
Ulyana-F1re: yes

articles/model-anatomy.adoc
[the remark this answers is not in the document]
Ulyana-F1re: I just never have seen short numbers used as ids in books or papers. `n1` is also meaningless. But your formatting of course helps to distinguish that these are ids. What if someone copies your text and shares it without formatting?Just to be sure, the keys that you use in the article are capitalized meaningful names like `MyConcept`, right?

articles/model-anatomy.adoc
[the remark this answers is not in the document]
Ulyana-F1re: So 'it' in the second part of the sentence is misleading.

articles/model-anatomy.adoc
Ulyana-F1re: Is INamed actually introduced for resolving references? I thought it has more generic use. 
Yes, a footnote will be better.

articles/model-anatomy.adoc
Ulyana-F1re: In the picture in section 2.9: annotated node is `X`, right? But `X` is not contained, it would be a parent. How then "contained" is equivalent to "annotated node"? I am confused now.

articles/model-anatomy.adoc
Ulyana-F1re: What is misleading here is `All these nodes still adhere to the intents of House's designer`. I guess you mean that these new (window) nodes do not interfere with the intents of House's designer, but also transport the additional information. "Not interfering" and "adhering" are not the same here.
Imho, the house example is confusing for the introduction of annotations. So that you need to write an extra excuse for it below.

articles/model-anatomy.adoc
Ulyana-F1re: I am missing the conclusion of this paragraph: it introduces the challenge/problem but doesn't get to a solution of LionWeb.

articles/model-anatomy.adoc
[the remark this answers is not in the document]
enikao: Not really, I'd consider `PetrolEngine` to be the name of that classifier. It _could_ also be the key. I'm thinking of the name as I chose names that implicitly should make some sense for the example without explaining all the context -- thus they're targeted at humans, not computers.

articles/model-anatomy.adoc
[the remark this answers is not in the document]
enikao: I was thinking of symmetry in this sense: https://medium.com/@donraab/symmetric-sympathy-2c59d4541d60

articles/model-anatomy.adoc
enikao: It's not introduced, and yes, it has more generic use -- for everything that has a name. I think we propose to use it as default behavior for ResoveInfo in some doc.

articles/model-anatomy.adoc
enikao: One day later, I'm not so sure about this paragraph in general -- it moves the needle towards formal spec.

articles/model-anatomy.adoc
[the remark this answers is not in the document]
dslmeinte: “Smalltalk is about big dreams. Java is about little realities.” 😃 (and true enough)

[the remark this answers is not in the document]
dslmeinte: As a mathematician, I'd also prefer to subscribe to the “looser” notion of symmetry – as in the blog that Niko mentioned – than the stricter, more geometrically-inspired one.

articles/model-anatomy.adoc
enikao: I came up with an example, does that help?

articles/model-anatomy.adoc
```diff
@@ -0,0 +1,678 @@
...
+:m3: ../metametamodel/metametamodel
+:refarch: ../reference-architecture/reference-architecture
+:roadmap: ../roadmap/roadmap
+= Anatomy of a LionWeb Model
+:toc: preamble
+:toclevels: 3
+:sectnums:
+
+This article walks us through a LionWeb model:
+What it is, what's inside, how the inside looks, what the contents mean, and why they are that way.
```
krinara86, line 12: NP(Nitpicking): What do "contents" mean here

articles/model-anatomy.adoc
```diff
@@ -0,0 +1,678 @@
...
+It is meant as an introduction to some of the details of LionWeb.
```
krinara86, line 13: Here maybe it helps to paint a picture in one or two words what the "some parts" mean. Is it technical or design level or user level. Maybe an "intended" audience. You do mention no knowledge is required but there are multipe perspectives of being involved in LionWeb.

articles/model-anatomy.adoc
```diff
@@ -0,0 +1,678 @@
...
+We don't assume any prior knowledge of LionWeb; some experience with general modeling ideas is helpful.
+
+== Introduction
+The fundamental world view in LionWeb is "everything is a model".
+This means we express everything we talk about in this context in the form of models.
+
+But what is a model?
+STOP -- don't run away!
+This question is infamous in the modeling community -- but we have to answer it given our world view as stated above.
```
krinara86, line 22: I like the disclaimer here. :-)  Is there some urls or blog entries that showcase some arguments or discussions about the word model within the model community?

articles/model-anatomy.adoc
```diff
@@ -0,0 +1,678 @@
...
+We deliberately don't use that term in our specifications, as it's so overloaded and ambiguous.
+For our purposes, we can always replace "model" with "a bunch of nodes".
+Then the question transforms to: What is a node?
+
+[[notation]]
+== Notation used in this article and diagrams
```
krinara86, line 28: I would expect the "Nodes" paragraph at this point after the question What is nodes. Also I agree I am not sure what the "notations" paragraph gains at this point.

articles/model-anatomy.adoc
```diff
@@ -0,0 +1,678 @@
...
+Boxes in diagrams for M2-level language elements are labelled only with the classifier, i.e. `Person`.
+Boxes that represent concepts or concept instances use `C` icon, and `@` icon for annotation (instance) boxes.
+
+[[nodes]]
+== Nodes
+A node is the fundamental building block in LionWeb.
+It consists of a unique identifier (**id**), a **classifier** to define the nature / type / class of the node, the node's **parent** node and lists of **properties**, **containments**, **references**, and **annotation instances**, respectively.
+We group a node's properties, containments, and references as the **features** of that node.
+
+Properties are simple values like `name: String`, `age: Integer` or `alive: Boolean`.
```
krinara86, line 40: Different font/ styling for LionWeb terminologies like Property, containment and References. It would also be nice to have these in bullet points or the likes:

* Properties are...
* Containment...
* References ...

articles/model-anatomy.adoc
```diff
@@ -0,0 +1,678 @@
...
+Containments are other nodes that this node is composed of; examples would be `mainEntrance: Door[1]` or `corners: Coordinate[0..\*]`.
+References are other nodes that this node relates to; examples would be `surface: Material[1]` or `stakeholders: Employee[1..*]`.
```
krinara86, line 42: A tree type diagram to describe the example here would be useful.

articles/model-anatomy.adoc
```diff
@@ -0,0 +1,678 @@
...
+It cannot be empty, and it must be unique for all nodes in the model.
+
+The valid character set coincides to the output of _base64url_-encoding.
+This comes very handy if we had to derive the node id from some other source, like a binary hash, or fully qualified name.
+
+We chose this character set, as it should be compatible with any run-time environment, and avoid any encoding issues.
+We don't limit the id's length, so any kind of other source can be represented -- fully qualified names tend to get very long, for example.
+
+We assume nodes with equal id string to be identical.
+Something went wrong if we had e.g. two different Java objects of class `Node`, but with equal `nodeId` field.
```
krinara86, line 59: The word "java" or even programming languages here come out of nowhere. Maybe the stage could be set even earlier with "What is Lionweb" and quickly describing traditional code and model driven DSL code in the first or the "Intended audience" section

articles/model-anatomy.adoc
```diff
@@ -0,0 +1,678 @@
...
+
+Node ids should NOT carry any meaning:
```
krinara86, line 61: This part is a bit confusing. For starters, I would assume ids are generally auto generated. 
Secondly, I like the idea of descriptive node ids for testing purposes. But, should it not be somehow made sure they are semantically not related to the nodes they are describing to avoid such confusions like "left" id for a node represeting the "right". If this situation will hardly happen, perhaps best to mention it. For me, I understood the use of this and the attitude that I should not read into their meanings when you said they are used for tests readability.

articles/model-anatomy.adoc
```diff
@@ -0,0 +1,678 @@
...
+Only because a node id happens to contain the string `frontLeftWheel`, we should NOT assume its classifier might be `Wheel`, or its containment might be `frontLeft`.
+These "descriptive" node ids are quite useful to read examples or in tests, as they provide more context.
+However, examples and tests are for human consumption -- real nodes (in their raw form) are not.
+
+[[meta-pointer]]
+=== Meta-pointer
+Before we get to the next part of a node, classifier, we need to discuss _meta-pointers_.
```
krinara86, line 68: I would avoid forward references (classifier) unless really needed. In this case, the definition of meta-pointer stands on its own and the classifier is anyway described as a kind of meta-pointer. So I dont feel this is necessary.

krinara86 (commented):
Review till meta-pointers

articles/model-anatomy.adoc
dslmeinte: Yeah, section 15.2 (pages 405-406) of my book explains this ;)

articles/model-anatomy.adoc
```diff
@@ -0,0 +1,678 @@
...
+The contained node makes no sense without its parent.
+Even if we kept the contained node, where would it go?
+We'd need to invent some special place for it.
+In contrast, the target of a reference is perfectly valid without the source.
+They can be deleted independently, and don't need to be deleted when sources pointing to them are deleted.
+
+Containments establish a tree shape for the model.
+With references, we regard the model as a graph, as a node can be target of several references.
+
+.Node graph example
```
erk-dik, line 474: This graph does not have any explanation and it is not refferred from anywhere in the text.

articles/model-anatomy.adoc
enikao: That's exactly the point I'm trying to make here. I've seen people having such assumptions, and want to debunk them.

articles/model-anatomy.adoc
```diff
@@ -0,0 +1,678 @@
...
+We need this triple, because a language can be uniquely identified by its key and version, and the element key must be unique within its language.
+With language version we support _language evolution_, i.e. the fact that real-world languages tend to change over time.
+
+We don't prescribe anything about language version, as different implementations might use different ways to represent language evolution.
+We just assume that equal strings refer to the identical version.
+
+=== Node classifier
+A node's classifier is a <<meta-pointer>> that defines the node's nature, or type or class -- "what kind of thing is this node?"
+A classifier of a node is equivalent to the class of a Java object.
+The classifier can either be a _concrete concept_ or an _annotation_. It cannot be an _abstract concept_ or _interface_, as we cannot instantiate them.
```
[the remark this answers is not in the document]
enikao: We're not talking about classifiers on their own, but the classifier of a node. As the node _is_ the instance, a node's classifier must be instantiable.

articles/model-anatomy.adoc
enikao: moved to appendix

articles/model-anatomy.adoc
```diff
@@ -0,0 +1,898 @@
...
+  a1 *-> an2: <<annotations>>
+}
+
+M2 -[hidden]- M1
+----
+
+An annotation that is applicable to all concepts says it annotates `Node` from LionWeb's <<builtins, built-in library>>.
+Annotations can _annotate_ other annotations; however, this approach introduces complexities and should be used with great caution.
+
+=== Differences between containment and annotation instance
```
dslmeinte, line 739: "instance**s**"?

dslmeinte (approved):

articles/model-anatomy.adoc
```diff
@@ -0,0 +1,898 @@
...
+It cannot be empty, and it must be unique for all nodes in the model.
+
+The valid character set coincides with the output of _base64url_-encoding.
+This comes very handy if we had to derive the node id from some other source, like a binary hash, or a fully qualified name.
+
+We chose this character set, as it should be compatible with any run-time environment, and avoid any encoding issues.
+We don't limit the id's length, so any kind of other source can be represented -- fully qualified names tend to get very long, for example.
+
+We assume nodes with equal id string to be identical.
+Assuming an implementation in an object-oriented language like Java, something went wrong if we had two different Java objects of class `Node`, but with equal `nodeId` field.
```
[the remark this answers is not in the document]
enikao: I somehow want to illustrate what the sentence before means. Any ideas how to do this better?

articles/model-anatomy.adoc
```diff
@@ -0,0 +1,898 @@
...
+For multiple references, we keep track which target is listed at which position.
+The same target can appear several times in a multiple reference.
+This can help, for example, to describe steps: `steps: Task[1..*] = {generate, saveAll, compile, saveAll}` refers to Task `saveAll` twice.
+
+LionWeb does not support bidirectional references.
+They are very hard to maintain, and also hard to process (i.e. if we want to traverse the whole model).
+
+[#resolveInfo]
+==== ResolveInfo
+For references, we store the target node's id, and additionally a `resolveInfo`.
```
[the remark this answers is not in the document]
enikao: Yes, one of the two can be null (that's mentioned below)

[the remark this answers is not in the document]
dslmeinte: Yeah, I think we decided that at one point, but it's not reflected in the specification: https://github.com/LionWeb-io/specification/blob/main/2023.1/serialization/serialization.adoc#reference

articles/model-anatomy.adoc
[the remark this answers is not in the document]
dslmeinte: Not without spending a lot more words on it, and honestly, it still wouldn't be illustrating. As soon as you run into intricacies like this, you're beyond this article.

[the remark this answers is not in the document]
enikao: removed

PR #325, opened 2025-01-10
Update ids for v 2024.1

ftomassetti:
We update lioncore.json and builtins.json to use the version-specific ID.
We also fix an issue in lioncore.json and builtins.json 2023.1 where the language version was set to 1 and not to 2023.1


dslmeinte (approved):

enikao (approved):

ftomassetti:
Thank you both!

Issue #382, opened 2025-07-07
ChangeReferenceResolveCommand has unneeded property "target"


joswarmer:
The `ChangeReferenceResolveInfo` command has an (optional) property `target`.

In general we add old values that are to be changed by a command as properties, but `target` isn't changed.
Therefore the `target` property should not be there.

Same for `DeleteReferenceResolveInfo`

enikao:
**Decision:** Remove unneeded `target`/`resolveInfo`.

We might add them (again) in the future if it turns out to be helpful.

PR #387, opened 2025-07-24
removed redundant target/resolveInfo part from messages

enikao:
https://github.com/LionWeb-io/specification/issues/382

### DIFF
--- a/delta/delta.schema.json
+++ b/delta/delta.schema.json
@@ -2078,9 +2078,6 @@
         "index": {
           "$ref": "#/$defs/index"
         },
-        "target": {
-          "$ref": "#/$defs/targetNode"
-        },
         "newResolveInfo": {
           "$ref": "#/$defs/resolveInfo"
         },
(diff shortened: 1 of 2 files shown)


delta/delta.schema.json
```diff
@@ -2096,13 +2093,12 @@
...
         "reference",
         "index",
-        "target",
         "newResolveInfo",
         "commandId"
       ],
       "additionalProperties": false,
-      "minProperties": 7,
-      "maxProperties": 8
+      "minProperties": 6,
```
dslmeinte, line 2100: Is it useful to constrain the number of properties like this, beyond specifying the `requiredProperties`?

dslmeinte (approved):

delta/delta.schema.json
enikao: As I understand JsonSchema, it only limits the possible space of values. Without that, we could add five times e.g. `newResolveInfo` (as JSON doesn't require unique "keys").

PR #390, opened 2025-07-28
Make suffixes of (technical names of) commands and events consistent

dslmeinte:


enikao (changes requested):
We would also need to update the JSON schema, and the examples in integration-testing.

This also shows the problem with the missing suffixes on `Composite` -- we now have the same technical name for two different messages, which violates our own statement in 3.6:

> Every kind of message has a unique technical name.

We could rename `NoOp`, but all other events have an `Event` suffix, so I'd go for uniformity.

dslmeinte:
> We would also need to update the JSON schema, and the examples in integration-testing.
> 
> This also shows the problem with the missing suffixes on `Composite` -- we now have the same technical name for two different messages, which violates our own statement in 3.6:
> 
> > Every kind of message has a unique technical name.
> 
> We could rename `NoOp`, but all other events have an `Event` suffix, so I'd go for uniformity.

Uniformity means: the technical name of each command has the suffix `Command`, and likewise `Event` for events?

enikao:
Ouch, I thought we already have the command / event suffixes -- but I was mistaken.
Shall we put this up for vote next Friday?

PR #392, opened 2025-08-01
Explain why the composite command and event have suffixes

dslmeinte:
Also: `NoOpEvent` -> `NoOp`

### DIFF
--- a/delta/commands.adoc
+++ b/delta/commands.adoc
@@ -1361,5 +1361,10 @@ We can nest composite commands arbitrarily.{fn-org360}
 .Technical name
 `CompositeCommand`
 
+****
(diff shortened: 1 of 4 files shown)


delta/commands.adoc
```diff
@@ -1361,5 +1361,10 @@ We can nest composite commands arbitrarily.{fn-org360}
 .Technical name
 `CompositeCommand`
 
+****
```
enikao, line 1364: I'm not sure we want this in the spec document.
According to https://diataxis.fr/, specs should not contain explanations.
I tend to put footnotes on these things with links to the explanations / rationale.

dslmeinte: I think a GH issue is just a bit too far removed. Given that this side note resolves potential confusion with the spec, I could even argue that it should be an actual part of it (regardless of https://diataxis.fr). So, I opt for a footnote.

enikao (approved):

PR #395, opened 2025-08-15
Fix name inconsistencies between events and commands

dslmeinte:


dslmeinte:
> Shall we also adjust the schema and the examples in integration repo?

Certainly! But not in this PR ;)

dslmeinte:
> Schema is in this repo

Fair point! 😄 I'll update that as well in a couple of minutes.

dslmeinte:
> There are 6 places changed in the spec, but only 3 in schema -- suspicious ...

Not at all: half of the change locations in the spec are references to the other half of the change locations in the spec.

enikao:
> > There are 6 places changed in the spec, but only 3 in schema -- suspicious ...
> 
> Not at all: half of the change locations in the spec are references to the other half of the change locations in the spec.

sorry, you're right -- too many lines in between.

PR #424, opened 2025-12-11
Cleanup and splitting huge files

enikao:
Mostly no content changes, just moving stuff around.

* Moved "Technical name" as first subheader in each message
  Rationale: We often search for the technical name, and it was very confusing that we had to look _above_ the found line (because "Technical name" used to be the last subheader in each message).

* Split up one file per query, sorted into one directory per query "category"
  Rationale: `queries.adoc` became too big and hard to find stuff.

* Split up one directory per command + event + diagram, sorted into one directory per command/event "category"
  Rationale: `commands.adoc` and `events.adoc` became too big and hard to find stuff.
  Command and corresponding event are in the same directory, as we often have to change them together.  

  Unfortunately, we have to copy `commandEventOverviewStyle.puml` everywhere, because PlantUML resolves includes relative to the _current_ file -- unlike AsciiDoctor. We include the diagrams in different contexts, so any relative path will fail.

Editorial content changes:
* Remove open questions
* add query parameter labels


delta/queries.adoc
```diff
@@ -9,6 +9,9 @@
 ===== Subscribe to partition changes
 This client wants to receive events on newly created (<<SubscribeToChangingPartitions.creation>> is true) or deleted (<<SubscribeToChangingPartitions.deletion>> is true) partitions, and automatically subscribe (<<SubscribeToChangingPartitions.partitions>> is true) to newly created partitions.{fn-org267}
 
+.Technical name
+`SubscribeToChangingPartitions`
+
```
joswarmer: For queries, we have two messages for each query,,the request and the response message. This document only mentions one technical name, so each implementation needs to decide how the two message are called. We already have two different choices in the current code base:
TypoeScript package:

- messagekind for requiest === <<technical name>>Request,  i.e. `SubscribeToPartitionContentsRequest`
- messagekind for reqponse === <<technicalname>>Response i.e. `SubscribeToPartitionContentsResponse`

And in the lionweb-server package:

- messagekind for request === <<technical name>> i.e. `SubscribeToPartitionContents`
- messagekind for response === <<technicalname>>Response, i.e. `SubscribeToPartitionContentsResponse`

It would be good to have explicit technical names for both the response and the request.

enikao: I see this as an implementation detail of the underlying protocol. If we used a protocol that explicitly supports synchronous calls, we'd use that feature.
Example: If we used http, there would not be any query response message, as the contents would be part of the http response to the query itself.

The naming in C# seems like a bug, but I think it already has been fixed: https://github.com/LionWeb-io/lionweb-csharp/blob/3c36a381e1913ac0aba3c0b69d0f72694067812e/src/LionWeb.Protocol.Delta/Message/Query/Query.cs#L107

Issue #425, opened 2025-12-12
Provide partition contents only on explicit request


enikao:
We don't want to overload clients with whole new partition trees on every new partition.

Thus:

1. We remove `SubscribeToChangingPartitions.partitions` parameter. Clients are NOT automatically subscribed to any new partitions.
2. `PartitionAdded` event contains only the partition node + its properties, not any containments / references / annotations

We accept that `AddPartition` and `PartitionAdded` are not symmetric any more (because `AddPartition` **does** contain a whole partition tree)

Relates to #267

enikao:
# Use cases for partitions
Both being informed about new ones live (`PartitionAdded`) and when listing existing ones.
Lists the required info about that partition.

* Does a partition with this id exist? 
  -> only node id
* Create a proxy for each partition
  -> node id, classifier
* Check if that partition is interesting for me based on classifier
  -> node id, classifier
* Check if that partition is interesting for me based on property values
  e.g. "I'm looking for the partition with name 'x'"
  -> node id, properties
* Check if that partition refers to my partition
  e.g. "Is this AuxiliaryPartition augmenting my base model with id 'x'?"
  -> node id, references
* Check if that partition is annotated by my annotation
  e.g. "Does this partition have a trace annotation pointing to my base model with id 'x'?"
  -> node id, annotations, annotation node

# Possible chunk contents

## Option A: Only node id

## Option B: Node id + classifier

## Option C: Node id + classifier + property

## Option D: Node id + classifier + property + references/containments/annotations

_excluding_ the contained/annotated nodes

## Option E: Everything

## Option F: Provide depth parameter

### F1: depth as in bulk protocol
i.e. omitted = everything, any number = levels of containments/annotations

### F2: depth as mentioned in option A/B/C/D/E

### F3: depth as mentioned in option A/B/C/D + E with number of levels

Issue #422, opened 2025-12-04
Clarify semantics of Replace (with new content) commands


enikao:
"Replace with new content" means there's an existing `toBeReplaced` node in the model, and we want to replace it with a _new_ node `replacement` node.

We support the following "replace with new content" commands:
* `ReplaceChild` / `ChildReplaced`
* `ReplaceAnnotation` / `AnnotationReplaced`

The `replacement` node can be a complete subtree. All nodes in that subtree must be _new_.

_new_ means the node MUST
* either not exist in the repository
* or be same or descendant of ` toBeReplaced`
  In this case, a client SHOULD keep the identity of the node in its internal representation (e.g. the C# or Java object).
  Rationale:
  * Enables "wrap" scenarios: `myParent.containment = new WrapNode { wrapped = myParent.containment }`
  * Keeps both "native" (e.g. Java object references) and LionWeb references to that node alive
  * Little chance of interference with other aspects, as all the "pseudo-new" nodes would be deleted anyways.    
  
  Question: Do we consider these nodes to be deleted + new, moved, or something else?

enikao:
Note: for `AddChild` and `AddAnnotation`, all _new_ nodes MUST NOT exist in the repository, i.e. they must be really new.
We don't have any ambiguity about keeping object identities alive there.

enikao:
Prime use case: Wrapping a node
Secondary use cases: Most refactorings 

Issues:
* If several nodes are reused and changed heavily, reconciling is complicated, and we don't have detailed deltas of these changes
* We might not be able to keep object identity if node concept changes

Alternatives for wrapping use case:
* Introduce "soft-delete" to keep object identity of wrapped node alive --> requires clean-up / garbage collection
* Keep wrapped node for some time --> Transaction
* Introduce "wrap" message --> Extremely specific

**Decision: go with approach outlined in the issue text above**

Issue #408, opened 2025-09-29
How to handle potentially large delta messages


enikao:
These delta messages might get large, because they can contain complete subtrees of nodes:

* `SubscribeToPartitionContentsResponse.contents`
* `ListPartitionsResponse.partitions`
* `AddPartition.newPartition` / `PartitionAdded.newPartition`
* `AddChild.newChild` / `ChildAdded.newChild`
* `ReplaceChild.newChild` / `ChildReplaced.newChild`
* `AddAnnotation.newAnnotation` / `AnnotationAdded.newAnnotation`
* `ReplaceAnnotation.newAnnotation` / `AnnotationReplaced.newAnnotation`

In #327 and #369, we've decided to allow complete subtrees (not only the node itself).

Potential problems with big messages:
* Might need lots of memory during (de)serialization
* Might clog the channel if the channel can only transmit one message at a time (i.e. all other messages have to wait until the big one is fully transmitted)
* Might exceed memory and/or processing time limits with some clients (especially browsers, seem to refuse WebSocket messages > 16 MB)

NOTE: Extreme cases like gigabytes of text in one property are out of scope. They can still lead to huge messages,
but don't represent a reasonable use case -- more a wrong choice of tools/infrastructure.

# Option A: Accept the risk, don't change anything

Pro:
* Already implemented
* Risk can be mitigated both in client/repo implementations and concrete use cases

Con:
* Build known flaw into delta protocol from the start

# Option B: Disallow deep subtrees, send more single messages instead

Pro:
* Aligns better with "stream of small messages" philosophy of delta protocol

Con:
* The problem discussed in #388 becomes even more prominent
* Requires dealing with unresolved references

## Variant Bx: Send whole subtree as one CompositeEvent with lots of delta events

Pro:
* Client knows it's fully initialized after processing the CompositeEvent

Con:
* Might still hit message size limits

# Option C: Split up big chunks in multiple messages

Pro:
* Can be applied similarly for bulk

Con:
* One logical message might be split up in several physical messages, complicating client implementations

# Option D: Handle big chunks via bulk

Pro:
* Protocol designed for that

Con:
* Need to synchronize between two protocols (no gap or duplication between bulk load and delta updates)

# Option E: Refer to special bulk endpoint for large chunks
In delta message refer to "special" bulk endpoint that contains current chunk.
Client uses bulk protocol to download chunk.
Repo sends more events based on first delta message above

(Git LFS works similar)

Pro:
* Works for git
[...]

PR #426, opened 2025-12-12
Specify split chunks, extend queries

enikao:
Implements #408, #422, #425, #428

### DIFF
--- a/delta/description.adoc
+++ b/delta/description.adoc
@@ -24,7 +24,7 @@ Authentication and authorization are out of scope of this API.
 
 [[channel, channel]]
 === Channel characteristics
-Logically, the channel has independent parts for each <<introduction, category>> of message;
+Logically, the channel has independent parts for each <<category>> of message;
 in practice, we might transmit some or all of them through a shared channel.
 
 This protocol is agnostic of its underlying transport or technical implementation.
@@ -89,7 +89,7 @@ The exceptions are <<splittableMessage, splittable messages>>.
 
 [[messageIdentity, message identity]]
 .Message identity
-Each message has some id unique to the <<participation>>:
+Each message has some id unique to the <<participation>> and <<category>>:
 
 * Queries have a <<queryIdType>> (for <<qry-ChunkedQueryResponse>> together with its <<continuedChunkSequenceNumberType>>).
 * Commands have a <<commandIdType>> (for <<cmd-ChunkedCommand>> together with its <<continuedChunkSequenceNumberType>>).{fn-org305}
--- a/delta/introduction.adoc
+++ b/delta/introduction.adoc
@@ -4,7 +4,7 @@ Delta API assumes a bidirectional communication channel between a <<client>> and
 The client initiates the channel with the repository to open a <<participation>>{fn-org337}.
 Both parties communicate through the established channel via _messages_.
 
-Delta API supports these message categories, loosely following CQRS patternfootnote:[https://en.wikipedia.org/wiki/Command_Query_Responsibility_Segregation]{fn-org291}:
+Delta API supports these message [[category, category]]categories, loosely following CQRS patternfootnote:[https://en.wikipedia.org/wiki/Command_Query_Responsibility_Segregation]{fn-org291}:
 
 * <<queries>> initiated/requested by the client, with synchronous response by the repository.
 A query requests some information from the repository without changing the repository's contents.


delta/commandsEvents/partitions/AddPartition/AddPartition.adoc
```diff
@@ -18,6 +18,9 @@ include::AddPartition-diagram.adoc[]
 Subtree to form a new partition.
 The _anchor node_ MUST NOT contain a _parent_ (as this is a partition, thus cannot have a parent).
 All nodes in that subtree MUST be <<new-nodes, new nodes>>.
++
+NOTE: This parameter differs from the <<PartitionAdded.newPartition, corresponding event parameter>>:
+The command parameter contains a whole subtree, whereas the event parameter only contains the partition node.
```
dslmeinte: I’d formulate it as "command**’s** parameter" and "event**’s** parameter".

delta/commandsEvents/partitions/AddPartition/PartitionAdded.adoc
```diff
@@ -9,7 +9,9 @@ This client is now subscribed to any changes to <<PartitionAdded.newPartition>>
 
 [horizontal]
 .Parameters
-[[PartitionAdded.newPartition, `newPartition`]]newPartition:: <<chunkType>>
+[[PartitionAdded.newPartition, `newPartition`]]newPartition:: <<shallowChunkType>>
+NOTE: This parameter differs from the <<AddPartition.newPartition, corresponding command parameter>>:
+The command parameter contains a whole subtree, whereas the event parameter only contains the partition node.
```
dslmeinte: See previous comment.

delta/messages.adoc
```diff
@@ -53,6 +56,19 @@ This allows a client to know when its command has been processed (either success
 
 [[integerType]]integer type:: An integer value.
 
+[[new-nodes, new nodes]]
+==== New nodes
+The term "new node" refers to nodes that MUST NOT yet exist in the repository.{fn-org327}
+
+[[replaced-nodes, replaced nodes]]
+==== Replaced nodes
+The term "replaced node" refers to nodes that are removed and re-added in the same command/event.{fn-org422}
```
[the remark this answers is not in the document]
enikao: Might this be confusing with all the `MoveXxx` commands?

delta/commandsEvents/children/ReplaceChild/ReplaceChild.adoc
```diff
@@ -21,6 +18,9 @@ include::ReplaceChild-diagram.adoc[]
 [horizontal]
 .Parameters
 [[ReplaceChild.newChild, `newChild`]]newChild:: <<chunkType>>
+Subtree to form a new child.
+The _anchor node_ MUST contain the <<ReplaceChild.parent>> as _parent_.{fn-org357}
+All nodes in that subtree MUST either be <<new-nodes>> or <<replaced-nodes>>.
```
joswarmer: Why also << replaced-nodes >>? Not even sure what << replaced-nodes >> means in this context.

joswarmer: My understanding: the spec used to state that all nodes must be new, now it says that they don't have to be new, but can also be replaced-nodes. This means that any node below `ReplaceChild.replacedChild` can be used as well.  

I cannot remember we discussed this (maybe I wasn't there), but why make this change?

delta/description.adoc
```diff
@@ -73,20 +73,51 @@ The repository also sends out the same events to subscribed clients, thus _not_
...
+
+Commands::
+* <<cmd-AddPartition>> <<AddPartition.newPartition>>
+* <<cmd-AddChild>> <<AddChild.newChild>>
+* <<cmd-ReplaceChild>> <<ReplaceChild.newChild>>
+* <<cmd-AddAnnotation>> <<AddAnnotation.parent>>
+* <<cmd-ReplaceAnnotation>> <<ReplaceAnnotation.newAnnotation>>
+
+Events::
+* <<evnt-PartitionAdded>> <<PartitionAdded.newPartition>>
```
joswarmer: Because of [https://github.com/LionWeb-io/specification/issues/425](https://github.com/LionWeb-io/specification/issues/425) this PartitionAdded chunk will always be one node, so it never becomes too big.

delta/continuedChunk/ChunkedCommand.adoc
```diff
@@ -0,0 +1,27 @@
...
+* <<cmd-ReplaceChild>> <<ReplaceChild.newChild>>
+* <<cmd-AddAnnotation>> <<AddAnnotation.parent>>
+* <<cmd-ReplaceAnnotation>> <<ReplaceAnnotation.newAnnotation>>
+
+.Technical name
+`ChunkedCommand`
+
+[horizontal]
+.Parameters
+[[ChunkedCommand.chunk, `chunk`]]chunk:: <<chunkType>> or <<shallowChunkType>>
```
joswarmer: I find it very bad practice to have a property with a join type. Also, I do not see why we need << shallowChunkType>> here. All chunks are full << chunkType >>'s

delta/commandsEvents/partitions/AddPartition/PartitionAdded.adoc
```diff
@@ -9,7 +9,9 @@ This client is now subscribed to any changes to <<PartitionAdded.newPartition>>
 
 [horizontal]
 .Parameters
-[[PartitionAdded.newPartition, `newPartition`]]newPartition:: <<chunkType>>
+[[PartitionAdded.newPartition, `newPartition`]]newPartition:: <<shallowChunkType>>
```
joswarmer: I see no reason to introduce a new type for this, we can (as we already did) use the << chunkType >> for this, with the explanation we already had, saying that the containments, references and annotations are empty.
Feels like overkill.

delta/description.adoc
```diff
@@ -73,20 +73,51 @@ The repository also sends out the same events to subscribed clients, thus _not_
 
 [[message, message]]
 === Message characteristics
-Each message is atomic; it cannot be split.
 
+[[messageAtomicity]]
+.Atomicity
+Most message is atomic; they cannot be split.
+The exceptions are all messages containing a <<chunkType>>.
```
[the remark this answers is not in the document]
enikao: I try to follow [Diataxis reference](https://diataxis.fr/reference/) guidelines, i.e. _not_ including rationale in this document. I use footnotes, ideally with issue links, for that.

delta/description.adoc
```diff
@@ -73,20 +73,51 @@ The repository also sends out the same events to subscribed clients, thus _not_
...
+* <<evnt-AnnotationAdded>> <<AnnotationAdded.newAnnotation>>
+* <<evnt-AnnotationReplaced>> <<AnnotationReplaced.newAnnotation>>
+
+[[messageIdentity, message identity]]
+.Message identity
 Each message has some id unique to the <<participation>>:
 
 * Queries have a _query id_.
 * Commands have a _command id_.{fn-org305}
 * Events have _command sources_.{fn-org306}
```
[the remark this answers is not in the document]
enikao: It's the same as for the corresponding query request, thus not really unique. Responses don't really exist on conceptual level (only for WebSocket binding).

delta/description.adoc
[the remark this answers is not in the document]
joswarmer: Ok, clear.

delta/commandsEvents/partitions/AddPartition/PartitionAdded.adoc
joswarmer: At all other places I found it used in the spec, I don't think it should be there, as I said  in several comments.

delta/continuedChunk/ChunkedCommand.adoc
joswarmer: See my comment at ListPartitions, I still see no need for the shallowchunkType there.
Also, I do not think that the number of partitions will ever be big enough to be a problem with the size of this message.

delta/description.adoc
```diff
@@ -73,20 +73,51 @@ The repository also sends out the same events to subscribed clients, thus _not_
...
 
+[[technicalName, technical name]]
+.Technical name
 Every kind of message has a unique _technical name_.
 The technical name is a valid programming language identifier as described for <<{m3}.adoc#IKeyed, IKeyed>>.
 Each binding to a specific protocol might use the technical name in a different way.
 For example, a JSON binding might have a member containing the technical name of the message, or a Protobuf binding might use the technical name as type name.
```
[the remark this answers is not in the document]
enikao: I'm not so sure. Let's say we use Protobuf for efficiency reasons. Then the technical name would become the Protobuf message name. Protobuf encodes the message name in some indexed form, so all our message kinds would fit in one byte. If we now forced the implementation to _also_ contain the text form of the technical name, We would inflate every message by a considerable amount.

delta/continuedChunk/ChunkedCommand.adoc
enikao: Let's say we implement Java AST in LW, and each class becomes a partition. Then every business system easily has tens of thousands of partitions. I think that warrants chunking.

delta/description.adoc
```diff
@@ -89,7 +90,7 @@ Each message has some id unique to the <<participation>>:
 
 [[technicalName, technical name]]
 .Technical name
-Every kind of message has a unique _technical name_.
+Every defined message has a unique _technical name_.
```
dslmeinte: Why not use “message category” instead of “defined message”?

delta/continuedChunk/ChunkedEvent.adoc
```diff
@@ -0,0 +1,27 @@
...
+* <<evnt-AnnotationReplaced>> <<AnnotationReplaced.newAnnotation>>
+
+.Technical name
+`ChunkedEvent`
+
+[horizontal]
+.Parameters
+[[ChunkedEvent.chunk, `chunk`]]chunk:: <<unconstrainedChunkType>>
+The continued chunk.
+MUST adhere to the same constraints as the related <<splittableMessage>>'s chunk.
```
joswarmer: To be precise: the complete chunk from all combined ChunkerEvents should adhere to the same constraints as the related <<splittableMessage>>'s chunk.
The individual ChunkedEvents are not constrained by this.

delta/messages.adoc
```diff
@@ -33,7 +33,34 @@ Same format as <<{m3}.adoc#identifiers, identifiers>>; must be unique per <<part
...
+[horizontal]
+[[singleChunkType, single chunk]]Single:::
+MUST contain a single _anchor node_, and all other nodes MUST be descendants of the _anchor node_.
+The chunk MUST contain all features of the _anchor node_ and all its descendants, i.e. the _anchor node_ is complete.
+
+[[multiChunkType, multi chunk]]Multi:::
+MAY contain more than one _root node_ (i.e. independent subtrees), and the chunk MUST contain all descendants of each _root node_.
+MUST contain all features of all contained nodes.
+
+[[shallowChunkType, shallow chunk]]Shallow:::
```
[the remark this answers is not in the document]
enikao: I don't get the difference -- there's only one mention of a shallow chunk:

image

joswarmer (approved):

Issue #430, opened 2026-01-07
Rename + unify ProtocolMessage?


enikao:
Both [delta](https://lionweb.io/specification/delta/delta-api.html#message) and [bulk](https://lionweb.io/specification/bulk/repo-access-api.html#messagesType) describe a `protocolMessage` / `message` field.
These are arbitrary messages that can be attached to any interchange between client and repository.

Shall we find a better name for them?
* They are named differently in bulk and delta, even though they serve the same purpose and have the same structure.
* In delta, "ProtocolMessage" distinguishes it from "Messages" in general (i.e. everything we send around), but that distinction is weak.

Maybe we can think about a new name based on their purpose: user-defined extensions, additional info, or attachments.

dslmeinte:
* CustomMetaData

enikao:
**Decision: `additionalInfos`**

dslmeinte:
Checked that we did this

Issue #421, opened 2025-11-28
How do ProtocolMessages of command/event, or queryRequest/queryResponse relate?


enikao:
If a repo gets a protocol message from a client, how should it deal with the message in the corresponding "reply"?
Should the repo copy the messages to the reply?

* For queryRequest messages, the reply is the associated queryResponse
* For command messages the reply is the associated event

# Option A: Unspecified
We don't make any statement about the relation.

# Option B: MUST repeat
The repo MUST copy any message.

Pro:
* Can use messages to communicate between clients

Con:
* Might transport lots of useless information

# Option C: SHOULD repeat
The repo SHOULD copy any message, but CAN omit messages from a "reply" if it has good reasons to do so (e.g. if the repo knows the message and processes it on its own).

# Option D: SHOULD NOT repeat
The repo SHOULD omit messages from the reply, but CAN copy messages if it has good reasons to do so (e.g. it knows this is a "broadcast" message)

# Option E: MUST NOT repeat
The repo MUST NOT copy any message

# Option F: Specify in message
Add a field to the message that specifies how the repo should deal with it.

joswarmer:
Personally I am in favor of the simplest one: `resend: boolean` only for commands, as this is the only one that is useful because:
- `resend` is not useful on `requests` as they will always result in a `response` for the requesting client only
- `resend` is not useful on `events` and `responses` as these are sent to clients, who never resend anything anyway
- `resend` is useful for `commands` as these may result in events to one or more (other) clients

PR #434, opened 2026-01-25
renamed, added spec for AdditionalInfo.distribute

enikao:
implements #421, #430

### DIFF
--- a/delta/description.adoc
+++ b/delta/description.adoc
@@ -93,30 +93,14 @@ Each message optionally can have one or more _additional infos_ consisting of:{f
 * [[additionalInfo.kind]]`kind` is an <<{m3}.adoc#identifiers, identifier>>-compatible string identifying the additional info type.
 Some kinds are pre-defined in this specification.
 A repository MAY reply with other, additional kinds.
-* [[additionalInfo.reach]]`reach` describes who should receive this additional info.{fn-org421}
-Its value MUST be one of:
-+
-[horizontal]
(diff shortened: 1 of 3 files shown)


delta/description.adoc
```diff
@@ -86,14 +86,21 @@ The technical name is a valid programming language identifier as described for <
...
-* [[protocolMessage.data]]`data` is a flat map with arbitrary keys and values.
-All values MUST be strings, the keys MUST be <<{m3}.adoc#identifiers, identifier>>-compatible.
+[[additionalInfo, additional info]]
+==== Additional Info
+Each message optionally can have one or more _additional infos_ consisting of:{fn-org331}{fn-org332}{fn-org430}
+
+* [[additionalInfo.kind]]`kind` is an <<{m3}.adoc#identifiers, identifier>>-compatible string identifying the additional info type.
+Some kinds are pre-defined in this specification.
+A repository MAY reply with other, additional kinds.
+* [[additionalInfo.distribute]]`distribute`? is an optional <<booleanType>> flag with default value `false`.
```
enikao: I went with `distribute` for the new flag.
Considered alternatives:
* `distributeToAllClients`
* `sendToAllParticipants`

`clients` or `participants` would be confusing: In _bulk_ we only talk about "clients", and in delta mostly about "participants". As we want to use AdditionalInfo in both specs, neither would work well.

I find "distribute" matches the expectations to the repository better than "resend".

PR #464, opened 2026-04-10
Fix small stuff in delta spec.

dslmeinte:
- add clarification
- add missing item
- fix layout

### DIFF
--- a/delta/description.adoc
+++ b/delta/description.adoc
@@ -117,8 +117,7 @@ All messages containing a <<unconstrainedChunkType>>.
 A client or repository CAN decide to split up a _splittable message_ if that message would exceed the acceptable message size.
 LionWeb does not define the acceptable message size, it SHOULD be a parameter for each implementation.
 
-All _splittable messages_ MAY contain a <<splitFlagType>>.
-(If present, it’s named `split`.)
+All _splittable messages_ MAY contain a <<splitFlagType>> — typically named `split`.
 If the flag is unset or `false`, the splittable message is self-contained.
 If the flag is `true`, one or more <<continuedChunk>> messages will follow.{fn-org425}
 
--- a/delta/description.adoc
+++ b/delta/description.adoc
@@ -118,6 +118,7 @@ A client or repository CAN decide to split up a _splittable message_ if that mes
 LionWeb does not define the acceptable message size, it SHOULD be a parameter for each implementation.
 
 All _splittable messages_ MAY contain a <<splitFlagType>>.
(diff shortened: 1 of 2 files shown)


erk-dik (approved):

delta/description.adoc
```diff
@@ -118,6 +118,7 @@ A client or repository CAN decide to split up a _splittable message_ if that mes
 LionWeb does not define the acceptable message size, it SHOULD be a parameter for each implementation.
 
 All _splittable messages_ MAY contain a <<splitFlagType>>.
+(If present, it’s named `split`.)
```
enikao: I'd not mention this here -- it's _very_ prone to maintenance issues.

delta/messages.adoc
```diff
@@ -36,6 +36,7 @@ Same format as <<{m3}.adoc#identifiers, identifiers>>; must be unique per <<part
 [[unconstrainedChunkType]]SerializationChunk:: Chunk of nodes as described in <<{serialization}.adoc, serialization specification>>, but without top-level `serializationFormatVersion` and `languages`  members.
 Thus, the serialization chunk has only the top-level member `nodes`.{fn-org358}
 
+****
```
enikao: This (rendered) box now separates the kinds of chunks from the chunk itself -- why?

dslmeinte: Aren’t these kinds of chunks essentially meant as a sub (description) list of the `SerializationChunk` type?

delta/description.adoc
dslmeinte: OK, will rephrase as a recommendation.

enikao: ("maintenance" as in: We'll mess up non-local changes when editing the document.)

delta/messages.adoc
enikao: Which they are: `SerializationChunk::` (two colons) vs. `Single:::` (three colons)

dslmeinte: That syntax doesn’t render correctly/meaningfully on most of the renderers I happen to be using.

enikao: If we wanted indentation I think this would be the correct way:

image

dslmeinte: If that syntax works more reliably than the triple colon syntax, then I’m for using `+-- ... --`. I don’t think indentation is required, but without either indentation or correct rendering of `:::`, it’s unclear where the sentences “Nodes in the chunk...” and “Chunks have different constraints...” really belong.

enikao (approved):

PR #465, opened 2026-04-10
Align ID and messageKind for NoOp event with technical name from the spec.

dslmeinte:
After propagation this solves https://github.com/LionWeb-io/lionweb-integration-testing/issues/99

### DIFF
--- a/delta/delta.schema.json
+++ b/delta/delta.schema.json
@@ -3384,10 +3384,10 @@
       "minProperties": 5,
       "maxProperties": 5
     },
-    "NoOpEvent": {
+    "NoOp": {
       "properties": {
         "messageKind": {
-          "const": "NoOpEvent"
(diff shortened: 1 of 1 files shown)


enikao (approved):
I don't have a strong opinion which side (spec or schema) to align with the other.

dslmeinte:
> I don't have a strong opinion which side (spec or schema) to align with the other.

btw: What was the reason again for "`ErrorEvent`" instead of "`Error`", also as the technical name? (The reason for "`Composite{Command|Event}`" is obvious.)

enikao:
> btw: What was the reason again for "`ErrorEvent`" instead of "`Error`", also as the technical name? (The reason for "`Composite{Command|Event}`" is obvious.)

There's also an `ErrorResponse` (as in: query response).

PR #468, opened 2026-04-17
Add old properties to move and replace commands

joswarmer:
Added properties to both the spec document and the SJON schema for the Move/MoveAndReplace commands.
I did ***not*** do a complete check for all commands to become similar to their events.

### DIFF
--- a/delta/errors.adoc
+++ b/delta/errors.adoc
@@ -54,7 +54,8 @@ Client provided a node with a parent, but the parent node does not exist.
 
 [[err-parentMismatch]]
 ==== The given parent of a node is not its parent 
-Client provided an existing node with a parent, but the parent does not have the node as a child (annotation or containment).
+Client provided an existing node N with a parent id in its parent property, but the node with this parent-id
+does not have the existing node N as a child (annotation or containment).
 
 .Technical name
 `parentMismatch`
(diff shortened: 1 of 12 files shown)


enikao:
Do we need to add some error results regarding inconsistent parameters on some of the messages?

joswarmer:
> Do we need to add some error results regarding inconsistent parameters on some of the messages?

Good point, I'll check them.
Done!  Quite a lot of additional errors.

delta/errors.adoc
```diff
@@ -45,6 +45,19 @@ Client provided an entry (either a node or a target/resolveInfo) and an index, b
...
+[[err-unknownParentNode]]
+==== The parent of a node does not exist 
+Client provided a node with a parent, but the parent node does not exist.
+
+.Technical name
+`unknownParentNode`
+
+[[err-parentMismatch]]
+==== The given parent of a node is not its parent 
+Client provided an existing node with a parent, but the parent does not have the node as a child (annotation or containment).
```
dslmeinte: This one’s a bit vague, IMHO. Something like: “Client provided an existing node with a declared parent, but that declared parent does not have the provided existing node as a child.”?

dslmeinte (approved):
But with 1 remark/question.

delta/errors.adoc
joswarmer: Right!. I'll clarify this better.

joswarmer: Client provided an existing node N with a parent id in its parent property, but the node with this parent-id
does not have the existing node N as a child (annotation or containment).

PR #470, opened 2026-04-29
Add some missing and changed details to delta specification

dslmeinte:


delta/description.adoc
```diff
@@ -151,16 +152,24 @@ If the flag is `true`, this is the last _continued chunk_.
 If the flag is `false`, one or more _continued chunk_ messages will follow.{fn-org425}
 
 There MUST NOT be any message in the same <<channel>> between the _chunked message_ and all related _continued chunks_.
+The continued chunks SHOULD be sent across the <<channel>> in increasing order of consecutive sequence number.
```
joswarmer: Since all messages MUST be sent in order, why make an exception for here.

PS I could not find the statement that all message MUST be sent in order, we should add this to the spec.

delta/description.adoc
```diff
@@ -170,9 +179,9 @@ It is only valid in <<commands>>, i.e. MUST NOT be used for <<events>> or <<quer
 If present and set to `true`, the repository MUST copy the additional info to all events that result from the command, and distribute the additional info to all participants receiving the event.
 
 * [[additionalInfo.message]]`message` is a human-readable string describing the additional info.
-* [[additionalInfo.data]]`data` is a flat map with 0 or more arbitrary keys and values.
+* [[additionalInfo.data]]`data` is a map – meaning, a dictionary-like JSON object – with 0 or more arbitrary keys and values.
```
joswarmer: If I understand correctly this is a (big) change in the actual JSON serialization.
What used to be:
```
additionalInfos: [ 
     {
        "kind": "info",
        "message": "Hello there!",
        "data": [ 
            { "key": "nodeid", "value: "Node-21" }
            { "key": "index", "value: "33" }
        ]
    },
]
```
now becomes:
```
additionalInfos: [
    {
        "kind": "info",
        "message": "Hello there!",
        "data": [ 
            { "nodeid": "Node-21" }
            { "index": "33" }
        ]
    },
]
```
Right?

dslmeinte: Correct.

The specification already always phrased it like that — at least in my interpretation. Now we just clarified it in the specification a bit, and aligned the JSON Schema with it.

delta/description.adoc
dslmeinte: I’m not sure we ever made that demand: the transport protocol might not be able to provide a guarantee about that. Maybe discuss that on Friday? It’ll then have to be without me, though.

delta/description.adoc
joswarmer: Still,  all implementations have to change now, just wanted to make sure I understand.

delta/description.adoc
joswarmer: To repeat a discussion t hat Meinte and I had in a TypeScript PR.

If events arrive out of order, even one single client will not work correctly using the delta protocol with the server. E.g. the client sends:

- `AddChild`: Add Child (anywhere) with node id = A
- `AddProperty`: Add property to node with id A
If these arrive not in order, and the `AddProperty` comes first, this command will be incorrect (node with ID A does not exist).

My conclusion is that if the order received is different from the order sent the whole delta protocol will, never work. Therefore we should specify that the transport protocol MUST deliver messages in the order that they are sent.

Or ... the messages from the client should have a sequence number (like the events), ensuring that the server knows their intended order.

Or .... any other ideas?

enikao: That's why we introduced [Event Sequence Number](https://lionweb.io/specification/delta/delta-api.html#event-sequence-number)

delta/description.adoc
enikao: I'd prefer not to mention JSON in the spec. I'm really fond of decoupling spec and serialization.

enikao: Add link to #471

delta/description.adoc
enikao: let's link, and use the right name of, `ContinuedChunkSequenceNumber`

delta/description.adoc
```diff
@@ -151,16 +152,24 @@ If the flag is `true`, this is the last _continued chunk_.
...
+The continued chunks SHOULD be sent across the <<channel>> in increasing order of consecutive sequence number.
 
 _Continued chunks_ include:
 
 * <<qry-ChunkedQueryResponse>>
 * <<cmd-ChunkedCommand>>
 * <<evnt-ChunkedEvent>>
 
+.Errors
+* Split message already completed: continued chunk received with a higher consecutive sequence number than the first one for which the <<continuedChunkCompletedFlagType>> was `true`.
```
enikao: let's link, and use the right name of, `ContinuedChunkSequenceNumber`

delta/description.adoc
```diff
@@ -151,16 +152,24 @@ If the flag is `true`, this is the last _continued chunk_.
...
+* Continued chunk sequence declared complete more than once: multiple continued chunks (with different consecutive sequence numbers) had their <<continuedChunkCompletedFlagType>> set to `true`.
```
enikao: let's link, and use the right name of, `ContinuedChunkSequenceNumber`

delta/description.adoc
```diff
@@ -151,16 +152,24 @@ If the flag is `true`, this is the last _continued chunk_.
...
+* Received continued chunk with consecutive sequence number that was already seen.
```
enikao: let's link, and use the right name of, `ContinuedChunkSequenceNumber`

delta/description.adoc
enikao: I'd rephrase to "... than a previous one for which ..."

delta/description.adoc
```diff
@@ -151,16 +152,24 @@ If the flag is `true`, this is the last _continued chunk_.
...
+* Continued chunks refers to a non-existent query/command/event.
```
enikao: chunk as singular

delta/description.adoc
```diff
@@ -151,16 +152,24 @@ If the flag is `true`, this is the last _continued chunk_.
...
+* Continued chunk refers to a query/command/event that’s not splittable or doesn’t have its <<splitFlagType>> set to `true`.
+
```
enikao, line 174: Another error: channel had a non-continued message before continued messages finished.

delta/continuedChunk/ChunkedEvent.adoc
enikao: Shall we rename this to "Continued Event" (and the same for "Continued Command" and "Continued query response"?

#472

enikao: What if `continuedChunkSequenceNumber` and `sequenceNumber` contradict (i.e. don't increase in the same way)?

See also #475

delta/description.adoc
joswarmer: As I remember we introduced event sequence number mainly to allow a client to see that a message was lost, not to allow delivery in the incorrect order. So the second option above is to add a CommandSequenceNumber.

As our main target we are using today is WebSockets and they guarantee delivery in order, I think we should require this from the network protocol (first option above). Note that allowing commands and events to arrive in random order makes things quite more complicated for both clients and servers.

delta/continuedChunk/ChunkedEvent.adoc
joswarmer: @enikao Good idea, sounds consistent

delta/description.adoc
enikao: You're right, that was the reason for event sequence number.

I also do the reordering in C# (I think that's not violating the spec).

For commands it makes less sense as the repo decides on the order anyways, so no client can be sure about the application order (of either its own commands or any other clients').

joswarmer: The server should always apply the commands in the order sent by one client.  For this it needs to know the order as intended. Even in the trivial example above this is needed.

The server _may_ decide to do some conflict resolution, between commands from different clients, but it will need the intended order as sent by the client to be able to make any kind of sensible decision.

Conflict resolution between commands _from the same client_ should _never_ be needed, as a client should send commands based on the status after having applied the previous command.

delta/continuedChunk/ChunkedEvent.adoc
dslmeinte: Let’s update the spec. in a different PR, after having discussed #475.

delta/description.adoc
dslmeinte: I think we should discuss this further outside of this PR: let’s talk about this next Friday (May 8th)?

dslmeinte:
Commits pushed which should address all review comments (at least the ones I marked as resolved).

delta/description.adoc
joswarmer: Agreed, important to discuss separately. I will try to be at the meeting this week, although it is an open office meeting.

joswarmer (approved):

delta/description.adoc
enikao: (sorry, forgot to unresolve)

dslmeinte: Is it OK to retain “dictionary“?

enikao: ok, although I don't see the clarification over "map" a few words in front.

dslmeinte: Pushed 1 more commit.

enikao (approved):

delta/description.adoc
enikao: See also #483, #484

Issue #469, opened 2026-04-20
MoveAndReplaceChildInSameContainment unclear / incorrect


joswarmer:
In  [https://lionWeb.io/specification/delta/delta-api.html#cmd-MoveAndReplaceChildInSameContainment](https://lionWeb.io/specification/delta/delta-api.html#cmd-MoveAndReplaceChildInSameContainment)  their is confusion what the `newIndex` exactly is, when the `oldIndex` < `newIndex`.

In the picture it shows that _After_, the `movedChild` is at `newIndex`.  However, it actually is at `newIndex -1` (just count the index in the picture). This is because the `movedChild` disappears from `oldIndex`, so the `replacedChild`'s index becomes one less.

Note that the index of `movedChild` is not always `newIndex`:
- `newIndex` if `oldIndex` > `newIndex`
- `newIndex` - 1 if `oldIndex` < `newIndex`

enikao:
# Option A: `NewIndex` in new list
Pro:
* `NewIndex` is always the index where the node will be afterwards (esp. for MoveAndReplace where `NewIndex` > `OldIndex`).
So a client can read the event like "ah, that's where the node is now".

Con:
* Very confusing

# Option B: `NewIndex` in old list
Pro:
* Easier to calculate `NewIndex`

Con:
* To figure out where the node is now, client needs to compare `OldIndex` and `NewIndex`, or search through the containment and find the index of the moved element.

# Option C: Relative offset
Pro:
* Independent of old / new list

Con:

# Option D: Relative node
Use `replacedNode` (for move-and-replace) or `previousNode` instead of index

Pro:
* Most local assumptions

Con:
* Edge case: move to position 0
* How to handle if `previousNode` is not there any more? Would need to remember its old position

Tentative Decision: Option C

We disallow offset 0
Negative offset means move left
Positive offset means move right
Offset that results outside list index is an error?

dslmeinte:
Another pro of C: it aligns extremely well with the pathological use cases for this delta, e.g. moving a child up/down 1 position.

dslmeinte:
> PS: `MoveAnnotationInSameParentCommand` and `MoveAndReplaceAnnotationInSameParentCommand` have the same issue.

Only `MoveAndReplaceAnnotationInSameParentCommand` though, I’d say.

PR #481, opened 2026-05-12
Rework move and replace <X> in same parent

dslmeinte:


joswarmer (approved):
Looks ok, assuming we do want the `newIndex` property in the event.

delta/commandsEvents/children/MoveAndReplaceChildInSameContainment/ChildMovedAndReplacedInSameContainment.adoc
```diff
@@ -1,39 +1,49 @@
...
-<<ChildMovedAndReplacedInSameContainment.replacedChild>>, and all its <<ChildMovedAndReplacedInSameContainment.replacedDescendants>>{fn-org370}, have been deleted.
+<<existing-nodes, Existing node>> <<ChildMovedAndReplacedInSameContainment.movedChild>> moved to replace <<existing-nodes, existing node>> <<ChildMovedAndReplacedInSameContainment.replacedChild>> inside <<ChildMovedAndReplacedInSameContainment.parent>>'s <<ChildMovedAndReplacedInSameContainment.containment>>.
+The <<ChildMovedAndReplacedInSameContainment.movedChild>> was previously at index <<ChildMovedAndReplacedInSameContainment.oldIndex>>), and the <<ChildMovedAndReplacedInSameContainment.replacedChild>> was previously at index (<<ChildMovedAndReplacedInSameContainment.oldIndex>> + <<ChildMovedAndReplacedInSameContainment.indexOffset>>).
```
enikao: superfluous closing parenthesis

delta/commandsEvents/children/MoveAndReplaceChildInSameContainment/ChildMovedAndReplacedInSameContainment.adoc
```diff
@@ -1,39 +1,49 @@
...
+The new index of the moved child equals <<ChildMovedAndReplacedInSameContainment.newIndex>>.
+<<ChildMovedAndReplacedInSameContainment.replacedChild>>, and all its <<ChildMovedAndReplacedInSameContainment.replacedDescendants>>{fn-org370}, have been deleted, WITHOUT changing references to any of the deleted nodes.
 
-If <<ChildMovedAndReplacedInSameContainment.oldIndex>> < <<ChildMovedAndReplacedInSameContainment.newIndex>>: All other children inside <<ChildMovedAndReplacedInSameContainment.parent>>'s <<ChildMovedAndReplacedInSameContainment.containment>> with previous index > <<ChildMovedAndReplacedInSameContainment.oldIndex>> have been moved to next lower index.
+In case <<ChildMovedAndReplacedInSameContainment.indexOffset>> > 0: all children inside <<ChildMovedAndReplacedInSameContainment.parent>>'s <<ChildMovedAndReplacedInSameContainment.containment>> whose indices previously were > <<ChildMovedAndReplacedInSameContainment.oldIndex>> but not equal to (<<ChildMovedAndReplacedInSameContainment.newIndex>> + 1), have been moved to the next lower index.
```
enikao: we refer to `newIndex` before we give the formula to it. Maybe reverse?

enikao: I was first puzzled by the "not equal to (`newIndex` + 1)" part, until I realized that's the replaced one. So I'd focus on that fact: "except the replaced node at (`newIndex` + 1)"

delta/commandsEvents/children/MoveAndReplaceChildInSameContainment/ChildMovedAndReplacedInSameContainment.adoc
```diff
@@ -1,39 +1,49 @@
...
+The value of <<ChildMovedAndReplacedInSameContainment.newIndex>> is calculated as <<ChildMovedAndReplacedInSameContainment.oldIndex>> + <<ChildMovedAndReplacedInSameContainment.indexOffset>> - 1.
```
enikao: I'd add parenthesis around every calculation.

delta/commandsEvents/children/MoveAndReplaceChildInSameContainment/ChildMovedAndReplacedInSameContainment.adoc
```diff
@@ -1,39 +1,49 @@
 [[evnt-ChildMovedAndReplacedInSameContainment]]
 ===== Child moved in same containment and replaced existing child
-<<existing-nodes, Existing node>> <<ChildMovedAndReplacedInSameContainment.movedChild>> (previously inside <<ChildMovedAndReplacedInSameContainment.parent>>'s <<ChildMovedAndReplacedInSameContainment.containment>> at <<ChildMovedAndReplacedInSameContainment.oldIndex>>) has replaced the <<existing-nodes, existing>> <<ChildMovedAndReplacedInSameContainment.replacedChild>> inside <<ChildMovedAndReplacedInSameContainment.parent>>'s <<ChildMovedAndReplacedInSameContainment.containment>> at <<ChildMovedAndReplacedInSameContainment.newIndex>>.
-<<ChildMovedAndReplacedInSameContainment.replacedChild>>, and all its <<ChildMovedAndReplacedInSameContainment.replacedDescendants>>{fn-org370}, have been deleted.
+<<existing-nodes, Existing node>> <<ChildMovedAndReplacedInSameContainment.movedChild>> moved to replace <<existing-nodes, existing node>> <<ChildMovedAndReplacedInSameContainment.replacedChild>> inside <<ChildMovedAndReplacedInSameContainment.parent>>'s <<ChildMovedAndReplacedInSameContainment.containment>>.
```
enikao: We should refer to #469 somewhere

delta/commandsEvents/children/MoveAndReplaceChildInSameContainment/ChildMovedAndReplacedInSameContainment.adoc
```diff
@@ -1,39 +1,49 @@
...
+MUST NOT be 0. +
+<<ChildMovedAndReplacedInSameContainment.oldIndex>> + <<ChildMovedAndReplacedInSameContainment.indexOffset>> MUST NOT be < 0, or >= the number of children inside <<ChildMovedAndReplacedInSameContainment.parent>>'s <<ChildMovedAndReplacedInSameContainment.containment>>.
+
 [[ChildMovedAndReplacedInSameContainment.replacedChild, `replacedChild`]]replacedChild:: <<targetNodeType>>
 
 [[ChildMovedAndReplacedInSameContainment.replacedDescendants, `replacedDescendants`]]replacedDescendants:: <<targetNodeType>>[]
 
+[[ChildMovedAndReplacedInSameContainment.newIndex, `newIndex`]]newIndex:: <<indexType>>
+MUST NOT be equal to <<ChildMovedAndReplacedInSameContainment.oldIndex>>. +
+MUST be equal to: <<ChildMovedAndReplacedInSameContainment.oldIndex>> + <<ChildMovedAndReplacedInSameContainment.indexOffset>> - &epsilon;, where &epsilon; = -1 if <<ChildMovedAndReplacedInSameContainment.indexOffset>> > 0 and = 0 if <<ChildMovedAndReplacedInSameContainment.indexOffset>> < 0.
```
enikao: I wouldn't start with greek symbols. We already have an "if" in there, then I'd just write out the whole two cases.

delta/commandsEvents/children/MoveAndReplaceChildInSameContainment/MoveAndReplaceChildInSameContainment-diagram.adoc
```diff
@@ -1,5 +1,5 @@
 |
-[plantuml,moveAndReplaceChildInSameContainment-before,svg]
```
enikao, line 2: The rendering does not match the intended content for me:

Image

delta/commandsEvents/children/MoveAndReplaceChildInSameContainment/MoveAndReplaceChildInSameContainment.adoc
```diff
@@ -1,17 +1,43 @@
...
+from index <<MoveAndReplaceChildInSameContainment.oldIndex>> to the target index <<MoveAndReplaceChildInSameContainment.oldIndex>> + <<MoveAndReplaceChildInSameContainment.indexOffset>>,
+replacing <<existing-nodes, current child>> <<MoveAndReplaceChildInSameContainment.replacedChild>>{fn-org371} inside <<MoveAndReplaceChildInSameContainment.containment>> at the target index,
+including all its descendants (including annotation instances).{fn-org469}
+<<MoveAndReplaceChildInSameContainment.replacedChild>> subsequently gets deleted.
+This does NOT change references to any of the deleted nodes.{fn-org285}
 The containment is inside <<MoveAndReplaceChildInSameContainment.parent>>.
 
-Delete <<existing-nodes, current child>> <<MoveAndReplaceChildInSameContainment.replacedChild>>{fn-org371} inside <<MoveAndReplaceChildInSameContainment.containment>> at <<MoveAndReplaceChildInSameContainment.newIndex>>, and all its descendants (including annotation instances).
-Does NOT change references to any of the deleted nodes.{fn-org285}
+The corresponding manipulation of a list of – implementation-level! – references to contained nodes can be described as a two-step process:
```
enikao: What about introducing a section about moving nodes in the same containment somewhere in section 3.7? We'll need a similar explanation for annotations.

delta/commandsEvents/children/MoveAndReplaceChildInSameContainment/MoveAndReplaceChildInSameContainment.adoc
```diff
@@ -1,17 +1,43 @@
...
+Now, the list contains one less element than before.
+
+This leaves two distinct scenarios depending on whether <<MoveAndReplaceChildInSameContainment.indexOffset>> is positive or negative.
+
+In the scenario with <<MoveAndReplaceChildInSameContainment.indexOffset>> > 0:
 
 [cols="a,a"]
 |===
 |Before |After
+include::MoveAndReplaceChildInSameContainment-diagram-offset-positive.adoc[]
```
enikao: Is it helpful to have two entirely different depictions how this should work (text example, diagram)? That's even harder to keep in sync and non-contradicting.

dslmeinte: I think it is because the indices afterwards are different for these scenarios.

delta/commandsEvents/children/MoveAndReplaceChildInSameContainment/MoveAndReplaceChildInSameContainment-diagram.adoc
dslmeinte: Please explain

delta/commandsEvents/children/MoveAndReplaceChildInSameContainment/MoveAndReplaceChildInSameContainment.adoc
dslmeinte: You mean specifically this paragraph?

delta/commandsEvents/children/MoveAndReplaceChildInSameContainment/MoveAndReplaceChildInSameContainment-diagram.adoc
enikao: This should be the example for positive offset. In the picture, the replaced node is _before_ the moved node, i.e. negative offset. Also, in the "after" picture, `movedChild` is _after_ `siblingC`.

delta/commandsEvents/children/MoveAndReplaceChildInSameContainment/MoveAndReplaceChildInSameContainment.adoc
enikao: The whole "how to handle indices when moving nodes in same containment" shenanigans.

delta/commandsEvents/children/MoveAndReplaceChildInSameContainment/MoveAndReplaceChildInSameContainment-diagram.adoc
dslmeinte: This is the wrong rendering!

delta/commandsEvents/children/MoveAndReplaceChildInSameContainment/MoveAndReplaceChildInSameContainment.adoc
enikao: Even in two variants: "move" and "move + replace"

dslmeinte: I’d say: section 3.8 “List manipulation characteristics”. OK?

delta/commandsEvents/children/MoveAndReplaceChildInSameContainment/MoveAndReplaceChildInSameContainment-diagram.adoc
enikao: This is what I see in `MoveAndReplaceChildInSameContainment.adoc`:

image

**Edit: wrong screenshot**

enikao: In `MoveChildInSameContainment-diagram.adoc` sibling `A`, `B`, `C` are in the correct order:

image

dslmeinte: Screenshot 2026-05-13 at 14 15 02
This is how I see it.

enikao: Looks better.

For the former case, can we introduce `siblingD`? So we see how indices behind the replaced node change.

dslmeinte: Can do that, but then would like to do that generically in section § 3.8.

dslmeinte: But do we want to change `MoveChildInSameContainment` to have `oldIndex` and `indexOffset` as well? In that case, there’s not really an ambiguity, I feel.

enikao: > But do we want to change `MoveChildInSameContainment` to have `oldIndex` and `indexOffset` as well? In that case, there’s not really an ambiguity, I feel.

I would do so. It's still easier to understand.

dslmeinte: Hmmm, that’s another change. I’m not sure whether that’s not a change too much.

joswarmer: I agree with Meinte, there is no ambiguity. I find both option easy to understand.

delta/description.adoc
```diff
@@ -190,3 +190,70 @@ All values MUST be strings.
...
+
+.Notation/shorthand
+We write _l_..._r_ for the set of integers _l_, (_l_ + 1), (_l_ + 2), ..., _r_; which includes _l_ and _r_ themselves.
+In effect: `oldIndex` MUST be in 0...(_n_ - 1).
+
+
+[[justMovingAnItem, moving an item (without replacing)]]
+==== Moving an item (without replacing)
+
+In addition to an `oldIndex`, we get passed a `newIndex`: an integer from 0...(_n_ - 1), that is NOT equal to `oldIndex` (as that would mean an idempotent action).
```
enikao: Please add links to the relevant commands

delta/description.adoc
```diff
@@ -190,3 +190,70 @@ All values MUST be strings.
...
+// `newIndex=0`: `*H[0]* _A[1] B[2] C[3] D[4] E[5] F[6] G[7]_` +
+
+// The item at index `oldIndex` is moved `indexOffset` indices forward/down/to the right (if `indexOffset` > 0) or backward/up/to the left (if `indexOffset` < 0), shifting other items out of way.
+// The result of moving the item at index `oldIndex` by `indexOffset` indices is the same as moving that item `indexOffset` times 1 index forward/down/to the right (if `indexOffset` > 0), or (-`indexOffset`) times 1 index backward/up/to the left (if `indexOffset` < 0).
+
+
+[[movingAndReplacing, moving an item and replacing another item]]
+==== Moving an item and replacing another item
+
+To avoid any potential ambiguities and misunderstanding, we decided to describe this scenario on the basis of an initial (“old”) situation and an _index offset_ that determines the target index.
```
enikao: Please add links to the relevant commands

delta/description.adoc
```diff
@@ -190,3 +190,70 @@ All values MUST be strings.
...
+
+[[manipulationInsideSingleList, manipulation inside a single list]]
+=== Characteristics of manipulation inside a single list
+
+There are two scenarios that involve manipulating a single list, and which benefit from some clarification with regard to how they work index-wise:
+
+1. Moving an item to a different index – the _target_ index – in a list, moving other list items aside as necessary.
+2. Moving an item to a different index – the _target_ index – in a list, *and* replacing the list item at that target index.
+
+For both scenarios, assume we have a list of _n_ > 1 items of some type.
```
enikao: I'd join this sentence and the next one, the separation doesn't simplify understanding. So "we have a list of n > 1 nodes". Not sure whether we should add something like "either children or annotations".

delta/description.adoc
```diff
@@ -190,3 +190,70 @@ All values MUST be strings.
...
+In our context, the items would be nodes, and the type one declared by a classifier’s feature.
+We also have an _old index_ `oldIndex`, which is an integer from 0 upto n-1.
```
enikao: I don't know the word "upto". I'd stay with simple words, e.g. "both inclusive"

enikao: Even easier: Only introduce `oldIndex` after the next paragraph, then we can use the introduced notation.

delta/description.adoc
```diff
@@ -190,3 +190,70 @@ All values MUST be strings.
...
+
+.Notation/shorthand
+We write _l_..._r_ for the set of integers _l_, (_l_ + 1), (_l_ + 2), ..., _r_; which includes _l_ and _r_ themselves.
```
enikao: Isn't the common way to write Ranges _two_ dots? Also distinguishes it from ellipsis.

enikao: I'd not use "l" and "r" (which possibly stands for "left" and "right")? Ideally use proper speaking variable names, in this case maybe "lower" and "upper".

delta/commandsEvents/children/MoveAndReplaceChildInSameContainment/MoveAndReplaceChildInSameContainment-diagram.adoc
enikao: The rendering is still off for me in the latest version (graphical sibling order is `C`, `A`, `B`).
My local rendering is exactly the same as online: https://www.plantuml.com/plantuml/uml/ZL4zJyCm4DtzAqvb0qM85LLHQRimLM9jCnZV9R7ad2lxYlZvX99hI0c8AJc_Z-TxNcMnIKvTC85g0OcGVDCAe4_an4SsnWMRCnnj8XQTidyDdfkdOUpY6-6czG7XPfSiDpJbhQVZHcsGiYD6ow8wHH8TWtKNqnT2HHOA7NXxaI_0ZfhjDzT4QAPUjqKvmnmYlhRiEJr9Ttm45rcivww73hoF0PyPwHqAUAWcYh66QytInH3J-JgbwfVtvyEocajT_vfwTvsgfVwXy9_VNLrdaX-tdgdwVqwLCSF2MkludCQC_r3iMuz8N8tll-gXKnS-0000

enikao: > I agree with Meinte, there is no ambiguity. I find both option easy to understand.

For me, the fundamental difference is:

# `oldIndex` and `newIndex`

1. Compare `oldIndex` and `newIndex`
2. If < 0: (do one thing) otherwise: (do something else)

Here I can screw up both steps:
In 1., I can compare them in two directions (either `oldIndex - newIndex` or `newIndex - oldIndex`)
In 2. I can do the wrong thing whether it's > 0 or < 0.

# `indexOffset`

1. If < 0: (do one thing) otherwise: (do something else)

Now I can screw up only one step, so half the solution space.

delta/description.adoc
dslmeinte: Ranges with 2 dots is “semi-standard”, it seems. But the distinction from ellipsis is a valid argument.

delta/commandsEvents/children/MoveAndReplaceChildInSameContainment/MoveAndReplaceChildInSameContainment-diagram.adoc
dslmeinte: > The rendering is still off for me in the latest version (graphical sibling order is `C`, `A`, `B`). My local rendering is exactly the same as online.

My local rendering was different — “interestingly” enough. I added some `[hidden]` arrows to try and force the order, and that seems to work after copying to the PlantUML playground. Can you check as well?

delta/moving/moving-and-replacing-diagram-offset-positive.adoc
```diff
@@ -1,5 +1,5 @@
 |
-[plantuml,moveAndReplaceChildInSameContainment-before,svg]
+[plantuml,moveAndReplaceChildInSameContainment-positive-offset-before,svg]
 ----
 !include commandEventOverviewStyle.puml
 left to right direction
```
enikao, line 6: `replacedChild` should be after `siblingC`, correct? My rendering shows it first, same as //www.plantuml.com/plantuml/png/ZL8zJyCm4DtzAwooWIB4YggeD5sOAh6s6OplabXopfLz87u-8N2DqmR4b5pUn_EzU9FOH7xhlVBOiEQWesixrk0YMdQ1bBBUfAHF9YAnJCczUaVjqy_OXtU4gdCUzCq-6cfnpjlEnqfi09D50aCiEa4YdhonPzCN145M0XrfTvQlj9aqkymr8gYvrsrHBZB7YEyj1yv0aXqVjGt4nb4_G4Sw10_F1FYX2_rGvugiXalDsWOVudYSKlJh-_5ZMI-bRdzDlRjCbLB_KBYlvxvf4lBrgQ-h_fzJhTIuy5Jhz3kLalPqSUWS05ATzwCcQaQOhekyB82LsWp2uSP-0W00

delta/commandsEvents/children/MoveAndReplaceChildInSameContainment/ChildMovedAndReplacedInSameContainment.adoc
```diff
@@ -1,35 +1,40 @@
...
+<<ChildMovedAndReplacedInSameContainment.replacedChild>>, and all its <<ChildMovedAndReplacedInSameContainment.replacedDescendants>>{fn-org370}, have been deleted, WITHOUT changing references to any of the deleted nodes.{fn-org469}
 
-If <<ChildMovedAndReplacedInSameContainment.oldIndex>> < <<ChildMovedAndReplacedInSameContainment.newIndex>>: All other children inside <<ChildMovedAndReplacedInSameContainment.parent>>'s <<ChildMovedAndReplacedInSameContainment.containment>> with previous index > <<ChildMovedAndReplacedInSameContainment.oldIndex>> have been moved to next lower index.
+In case <<ChildMovedAndReplacedInSameContainment.indexOffset>> > 0: all children inside <<ChildMovedAndReplacedInSameContainment.parent>>'s <<ChildMovedAndReplacedInSameContainment.containment>> whose indices previously were > <<ChildMovedAndReplacedInSameContainment.oldIndex>> have been moved to the next lower index, _except_ the one that was previously at index (<<ChildMovedAndReplacedInSameContainment.oldIndex>> + <<ChildMovedAndReplacedInSameContainment.indexOffset>>).
```
enikao: > except the one that was previously at index (oldIndex + indexOffset)

That's always the replaced one, correct? I'd add that.

delta/commandsEvents/children/MoveAndReplaceChildInSameContainment/ChildMovedAndReplacedInSameContainment.adoc
```diff
@@ -1,35 +1,40 @@
...
+The moved child now resides at the index (<<ChildMovedAndReplacedInSameContainment.oldIndex>> + <<ChildMovedAndReplacedInSameContainment.indexOffset>> - 1).
 Example: +
```
enikao: I'd add that we're replacing `F` with `X`

delta/commandsEvents/children/MoveAndReplaceChildInSameContainment/ChildMovedAndReplacedInSameContainment.adoc
```diff
@@ -1,35 +1,40 @@
...
-`oldIndex=3`: `A[0] B[1] C[2] *X[3]* D[4] E[5] F[6] G[7]` -> +
-`newIndex=5`: `A[0] B[1] C[2] _D[3] E[4]_ *X[5]* _G[6]_`
+`oldIndex=3`:    `A[0] B[1] C[2] *X[3]* D[4] E[5] F[6] G[7]` -> +
+`indexOffset=3`: `A[0] B[1] C[2] _D[3] E[4]_ *X[5]* _G[6]_` +
+`newIndex=5`
 
-If <<ChildMovedAndReplacedInSameContainment.oldIndex>> > <<ChildMovedAndReplacedInSameContainment.newIndex>>: All other children inside <<ChildMovedAndReplacedInSameContainment.parent>>'s <<ChildMovedAndReplacedInSameContainment.containment>> with previous index >= <<ChildMovedAndReplacedInSameContainment.oldIndex>> have been moved to next lower index.
+In case <<ChildMovedAndReplacedInSameContainment.indexOffset>> <  0: all children inside <<ChildMovedAndReplacedInSameContainment.parent>>'s <<ChildMovedAndReplacedInSameContainment.containment>> whose indices previously were > <<ChildMovedAndReplacedInSameContainment.oldIndex>> have been moved to next lower index.
+The moved child now resides at the index (<<ChildMovedAndReplacedInSameContainment.oldIndex>> + <<ChildMovedAndReplacedInSameContainment.indexOffset>>).
 Example: +
```
enikao: I'd add that we're replacing `B` with `X`

delta/commandsEvents/children/MoveAndReplaceChildInSameContainment/ChildMovedAndReplacedInSameContainment.adoc
```diff
@@ -1,35 +1,40 @@
 [[evnt-ChildMovedAndReplacedInSameContainment]]
 ===== Child moved in same containment and replaced existing child
```
enikao, line 2: Maybe split the lines with a few more hard-breaks. Yes, every sentence is a line, but with all the links the lines get pretty long.

delta/commandsEvents/children/MoveAndReplaceChildInSameContainment/MoveAndReplaceChildInSameContainment.adoc
```diff
@@ -1,17 +1,43 @@
 [[cmd-MoveAndReplaceChildInSameContainment]]
 ===== Move child in same containment and replace existing child
-Move <<existing-nodes, existing node>> <<MoveAndReplaceChildInSameContainment.movedChild>> within <<MoveAndReplaceChildInSameContainment.containment>> from <<MoveAndReplaceChildInSameContainment.oldIndex>> to <<MoveAndReplaceChildInSameContainment.newIndex>>.
+Move <<existing-nodes, existing node>> <<MoveAndReplaceChildInSameContainment.movedChild>> within <<MoveAndReplaceChildInSameContainment.containment>>
+from index <<MoveAndReplaceChildInSameContainment.oldIndex>> to the target index (<<MoveAndReplaceChildInSameContainment.oldIndex>> + <<MoveAndReplaceChildInSameContainment.indexOffset>>),
+replacing <<existing-nodes, current child>> <<MoveAndReplaceChildInSameContainment.replacedChild>>{fn-org371} inside <<MoveAndReplaceChildInSameContainment.containment>> at the target index,
+including all its descendants (including annotation instances).{fn-org469}
```
enikao: Maybe move to the next sentence?

delta/commandsEvents/children/MoveAndReplaceChildInSameContainment/MoveAndReplaceChildInSameContainment.adoc
```diff
@@ -1,17 +1,43 @@
...
+<<MoveAndReplaceChildInSameContainment.replacedChild>> subsequently gets deleted.
+This does NOT change references to any of the deleted nodes.{fn-org285}
 The containment is inside <<MoveAndReplaceChildInSameContainment.parent>>.
 
-Delete <<existing-nodes, current child>> <<MoveAndReplaceChildInSameContainment.replacedChild>>{fn-org371} inside <<MoveAndReplaceChildInSameContainment.containment>> at <<MoveAndReplaceChildInSameContainment.newIndex>>, and all its descendants (including annotation instances).
-Does NOT change references to any of the deleted nodes.{fn-org285}
+The corresponding manipulation of a list of – implementation-level! – references to contained nodes can be described as a two-step process:
```
enikao: Can we avoid the term "references" here? It's confusing in context of a containment.

dslmeinte: The thing is: these are fundamentally language-level references to objects, without any parent-child semantics involved. The parent-child semantics are emergent from the correctness of the algorithm. If I don’t make that distinction explicitly, it could become confusing because people might try and draw in the parent-child semantics _during_ this process.

delta/commandsEvents/children/MoveAndReplaceChildInSameContainment/ChildMovedAndReplacedInSameContainment.adoc
dslmeinte: You mean wrapping breaks (soft wraps)?

delta/moving/moving-and-replacing-diagram-offset-positive.adoc
dslmeinte: Ah wait, can add another `[hidden]` marking to fix.

delta/commandsEvents/children/MoveAndReplaceChildInSameContainment/ChildMovedAndReplacedInSameContainment.adoc
enikao: Pressing enter a few times so we can read it without soft-wrap and attach comments to the proper place.

delta/commandsEvents/children/MoveAndReplaceChildInSameContainment/MoveAndReplaceChildInSameContainment.adoc
enikao: What about "pointer"? Or maybe we don't need it at all -- just talk about "list of contained nodes"?

delta/commandsEvents/children/MoveAndReplaceChildInSameContainment/ChildMovedAndReplacedInSameContainment.adoc
dslmeinte: What do you mean with “attach comments to the proper place”?

enikao: If one sentence spans 10 lines in GH review, it's harder to attach a review comment to the relevant piece (as the comment is per-physical-line).

dslmeinte: Ah, _those_ comments :)

delta/commandsEvents/children/MoveAndReplaceChildInSameContainment/MoveAndReplaceChildInSameContainment-diagram.adoc
enikao: Diagrams are looking good now.

delta/description.adoc
```diff
@@ -190,3 +190,75 @@ All values MUST be strings.
...
+We describe this scenario as a two-step process:
+
+1. The list’s item with index `targetIndex` is set to item with initial index `oldIndex`, overwriting and effectively replacing the item to replace, which gets deleted as a result.
+Now, the list contains the same number of items as before, just with two occurrences of the item with initial inde `oldIndex`, namely at indices `oldIndex` and `targetIndex`.
+2. The item with index `oldIndex` is removed from the list, with the items after it being moved to the next lower index — this includes the item moved in step 1 itself if (`indexOffset` > 0).
+Now, the list contains one less item than before.
+
+For clarity, we can discuss two distinct “sub” scenarios depending on whether `indexOffset` is positive or negative.
+
+*TODO*
```
enikao: I guess we move the details from the other places here and refer to it?

dslmeinte: Correct — working on that now

delta/commandsEvents/annotations/MoveAnnotationInSameParent/AnnotationMovedInSameParent.adoc
```diff
@@ -1,32 +1,26 @@
 [[evnt-AnnotationMovedInSameParent]]
 ===== Annotation moved in same parent
-<<existing-nodes, Existing node>> <<AnnotationMovedInSameParent.movedAnnotation>> (previously inside <<AnnotationMovedInSameParent.parent>>'s annotations at <<AnnotationMovedInSameParent.oldIndex>>) has been moved inside <<AnnotationMovedInSameParent.parent>>'s annotations at <<AnnotationMovedInSameParent.newIndex>>.
+<<existing-nodes, Existing node>> <<MoveAnnotationInSameParent.movedAnnotation>>
```
enikao: All these references point to the command

delta/commandsEvents/children/MoveChildInSameContainment/ChildMovedInSameContainment.adoc
```diff
@@ -1,35 +1,29 @@
 [[evnt-ChildMovedInSameContainment]]
 ===== Child moved in same containment
-<<existing-nodes, Existing node>> <<ChildMovedInSameContainment.movedChild>> (previously inside <<ChildMovedInSameContainment.parent>>'s <<ChildMovedInSameContainment.containment>> at <<ChildMovedInSameContainment.oldIndex>>) has been moved
-inside <<ChildMovedInSameContainment.parent>>'s <<ChildMovedInSameContainment.containment>> at <<ChildMovedInSameContainment.newIndex>>.
+<<existing-nodes, Existing node>> <<MoveChildInSameContainment.movedChild>>
```
enikao: All these references point to the command

delta/description.adoc
```diff
@@ -190,3 +190,92 @@ All values MUST be strings.
...
 This additional info is symmetric to <<{bulk}.adoc#additionalInfo, additional infos in bulk API>>.
+
+
+[[manipulationInsideSingleList, manipulation inside a single list]]
+=== Characteristics of manipulation of a list
+
+A number of commands and events deal with moving a node in a single listfootnote:[We assume all our lists to be ordered.], and possibly replacing another node in the same list.
+Such a list will either consist of the child nodes of a <<containment>> on a `parent` node, or of the annotation nodes on a `parent` node.
+More precisely: such a list consists of (language-level) _pointers_ to nodes.
+More than with commands and events that involve moving a node from one list to another, there’s chance for confusion with regard to the indices before and after completion of the intended operation.
```
enikao: I had to read this sentence 3 times to understand it, maybe we can simplify it?

delta/description.adoc
```diff
@@ -190,3 +190,92 @@ All values MUST be strings.
...
+
+There are two types of operations that involve manipulating a single list, and which benefit from clarification with regard to how they work index-wise:
+
+1. Moving a node to a different index – the _target_ index – in a list, moving other list nodes aside as necessary.
+2. Moving a node to a different index – the _target_ index – in a list, *and* replacing the list node at that target index.
+Note that for this operation, the moved node might not reside at that target index afterward, because of the replacing!
+
+To avoid ambiguities in the description of such operations, and reduce the chance of misunderstanding the specification as much as possible, we chose to phrase such operations – “in relative coordinates” – in terms of an _index offset_.
```
enikao: Why both dashes and quotes formatting?

delta/description.adoc
```diff
@@ -190,3 +190,92 @@ All values MUST be strings.
...
+
+For convenience, we first establish some notation/shorthand:
+
+.Integer ranges
+We write _lower_.._upper_ for the set of integers consisting of: _lower_, (_lower_ + 1), (_lower_ + 2), ..., _upper_.
+The *integer range* _lower_.._upper_ includes _lower_ and _upper_ themselves.
+
+For both scenarios above, assume we have a list of _n_ > 1 nodes.
+We also have an _old index_ `oldIndex`, which MUST be an integer in 0..(_n_ - 1): the (unique) index that the node to move appears in the list.
```
enikao: Shall we also establish the notations "node to move" and "node to replace"?

delta/description.adoc
```diff
@@ -190,3 +190,92 @@ All values MUST be strings.
...
+`indexOffset` can be either positive or negative, meaning the movement will be towards higher indices (positive), or towards lower indices (negative).
+
+
+[[justMovingANode, moving a node (without replacing)]]
+==== Moving a node (without replacing)
+
+This pertains to the following commands and events:
+<<cmd-MoveChildInSameContainment>>, <<cmd-MoveAnnotationInSameParent>>, <<evnt-ChildMovedInSameContainment>>, <<evnt-AnnotationMovedInSameParent>>.
+
+Set `newIndex` = (`oldIndex` + `indexOffset`).
```
enikao: is `newIndex` the same as _target index_ above?

delta/description.adoc
```diff
@@ -190,3 +190,92 @@ All values MUST be strings.
...
+Graphically:
+
+For `indexOffset` > 0:
+[cols="a,a","a"]
+|===
+|Before |After
+include::moving/moving-diagram-offset-positive.adoc[]
+|===
+
+Example with `indexOffset` = 2 > 0): +
```
enikao: > Example _moving `X`_ with

delta/description.adoc
dslmeinte: For moving without replacing: yes; for moving with replacing: no. That’s why I want to talk about “target index” in general. For just moving I can immediately specialize it to “new index”.

delta/commandsEvents/annotations/MoveAnnotationInSameParent/AnnotationMovedInSameParent.adoc
```diff
@@ -1,32 +1,26 @@
...
-`oldIndex=3`: `A[0] B[1] C[2] *X[3]* D[4] E[5] F[6]` -> +
-`newIndex=5`: `A[0] B[1] C[2] _D[3] E[4]_ *X[5]* F[6]`
-
-If <<AnnotationMovedInSameParent.oldIndex>> > <<AnnotationMovedInSameParent.newIndex>>: All other annotations inside <<AnnotationMovedInSameParent.parent>>'s annotations with previous index >= <<AnnotationMovedInSameParent.newIndex>> and previous index < <<AnnotationMovedInSameParent.oldIndex>> have been moved to next lower index.
-Example: +
-`oldIndex=3`: `A[0] B[1] C[2] *X[3]* D[4] E[5] F[6]` -> +
-`newIndex=1`: `A[0] *X[1]* _B[2] C[3]_ D[4] E[5] F[6]`
-
-<<AnnotationMovedInSameParent.oldIndex>> MUST NOT be equal to <<AnnotationMovedInSameParent.newIndex>>.
+This conforms to <<justMovingANode, the action of moving a node in a single list>> — see that section for more information, diagrams, and examples.
```
enikao: We use the term "single" as the opposite of "multiple", so maybe it's confusing here.

delta/commandsEvents/annotations/MoveAnnotationInSameParent/MoveAnnotationInSameParent.adoc
```diff
@@ -1,16 +1,12 @@
...
+by <<MoveAnnotationInSameParent.indexOffset>> indices
+to index (<<MoveAnnotationInSameParent.oldIndex>> + <<MoveAnnotationInSameParent.indexOffset>>).
 
-[cols="a,a"]
-|===
-|Before |After
-
-include::MoveAnnotationInSameParent-diagram.adoc[]
-|===
+This conforms to <<justMovingANode, the action of moving a node in a single list>> — see that section for more information, diagrams, and examples.
```
enikao: Same "single" vs. "multiple" issue

delta/commandsEvents/annotations/MoveAnnotationInSameParent/MoveAnnotationInSameParent.adoc
```diff
@@ -19,23 +15,27 @@ include::MoveAnnotationInSameParent-diagram.adoc[]
...
 [[MoveAnnotationInSameParent.movedAnnotation, `movedAnnotation`]]movedAnnotation:: <<targetNodeType>>
 
 [[MoveAnnotationInSameParent.commandId]]commandId:: <<commandIdType>>
 
 [[MoveAnnotationInSameParent.additionalInfos]]additionalInfos:: <<additionalInfosType>>
 
 .Errors
-* <<err-unknownIndex>> for <<MoveAnnotationInSameParent.newIndex>> (is beyond (the number of annotations of <<MoveAnnotationInSameParent.parent>> - 1)
-* <<err-unknownIndex>> <<MoveAnnotationInSameParent.movedAnnotation>> is not at  <<MoveAnnotationInSameParent.oldIndex>> of <<MoveAnnotationInSameParent.parent>>
+* <<err-unknownIndex>> for <<MoveAnnotationInSameParent.oldIndex>> is not an integer in the set 0..(the number of annotations on <<MoveChildInSameContainment.parent>> - 1) — see <<justMovingANode>> for an explanation of the `left`..`right` notation
```
enikao: I think the errors are also the same for all <<JustMovingANode>>, so I'd tend to list them there (or only additional ones here).

delta/description.adoc
```diff
@@ -190,3 +190,94 @@ All values MUST be strings.
...
 
 This additional info is symmetric to <<{bulk}.adoc#additionalInfo, additional infos in bulk API>>.
+
+
+[[manipulationInsideSingleList, manipulation inside a single list]]
+=== Characteristics of manipulation of a list
+
+A number of commands and events deal with moving a node in a single listfootnote:[We assume all our lists to be ordered.], and possibly replacing another node in the same list.
+Such a list will either consist of the child nodes of a <<containment>> on a `parent` node, or of the annotation nodes on a `parent` node.
+More precisely: such a list consists of (language-level) _pointers_ to nodes.
```
enikao: "language" is overloaded. Maybe "implementation language" or "host language"?

delta/description.adoc
```diff
@@ -190,3 +190,94 @@ All values MUST be strings.
...
+`indexOffset` can be either positive or negative, meaning the movement will be towards higher indices (positive), or towards lower indices (negative).
+
+
+[[justMovingANode, moving a node (without replacing)]]
+==== Moving a node (without replacing)
+
+This pertains to the following commands and events:
+<<cmd-MoveChildInSameContainment>>, <<cmd-MoveAnnotationInSameParent>>, <<evnt-ChildMovedInSameContainment>>, <<evnt-AnnotationMovedInSameParent>>.
+
+Set `newIndex` to the target index (`oldIndex` + `indexOffset`).
```
enikao: Formally correct, but I find it confusing that everything else besides "target index" is formatted. Maybe set it italic?

delta/description.adoc
```diff
@@ -190,3 +190,94 @@ All values MUST be strings.
...
+For convenience, we first establish some notation/shorthand:
+
+.Integer ranges
+We write _lower_.._upper_ for the set of integers consisting of: _lower_, (_lower_ + 1), (_lower_ + 2), ..., _upper_.
+The *integer range* _lower_.._upper_ includes _lower_ and _upper_ themselves.
+
+For both scenarios above, assume we have a list of _n_ > 1 nodes.
+We also have an _old index_ `oldIndex`, which MUST be an integer in 0..(_n_ - 1): the (unique) index that the *node to move* appears at in the list.
+Finally, we have an _index offset_ `indexOffset`, which MUST be an integer unequal to 0, and such that the _target index_ (`oldIndex` + `indexOffset`) is in 0..(_n_ - 1).
+`indexOffset` can be either positive or negative, meaning the movement will be towards higher indices (positive), or towards lower indices (negative).
```
enikao: I'd talk about "positions" instead of "indices", but that might be personal preference.

delta/description.adoc
```diff
@@ -190,3 +190,94 @@ All values MUST be strings.
...
+|===
+|Before |After
+include::moving/moving-diagram-offset-negative.adoc[]
+|===
+
+Example moving `X` with `indexOffset` = -2 < 0: +
+`oldIndex=3`: `A[0] B[1] C[2] *X[3]* D[4] E[5] F[6] G[7]` -> +
+`newIndex=1`: `A[0] *X[1]* _B[2] C[3]_ D[4] E[5] F[6] G[7]` +
+
+In both cases, only the nodes with initial indices in min(`oldIndex`, `newIndex`)..max(`oldIndex`, `newIndex`) are affected.
```
enikao: > min(`oldIndex`, `newIndex`)..max(`oldIndex`, `newIndex`)

This needs advanced parsing to figure out which text is part of the range. Maybe format all ranges completely as backticks?

delta/commandsEvents/annotations/MoveAnnotationInSameParent/AnnotationMovedInSameParent.adoc
dslmeinte: "a single" &rarr; "one specific" OK?

delta/description.adoc
dslmeinte: I’ve typeset "min" and "max" to be bold: would that already help? To me, having everything in one font doesn’t make it more readable.

delta/description.adoc
dslmeinte: Just here – as in: “higher/lower positions” –, or in general?

delta/description.adoc
dslmeinte: Text was changed a bit in the meanwhile, but I’ll set "target index" italic.

delta/commandsEvents/annotations/MoveAnnotationInSameParent/MoveAnnotationInSameParent.adoc
dslmeinte: I’m not sure: in a good API spec., I’d expect all errors to be listed in such a way that I don’t have to dereference. And some errors are specific to the message itself, and not to moving (and replacing) in a list at all.

enikao: Also true. Let's see how many inconsistencies we'll produce (-:

delta/description.adoc
enikao: "Index" is for me about the label we give a specific slot. "Position" is about adjacent slots.
So: I'd keep the terms "IndexOffset" etc, but talk about positions if it's not about the label.

delta/description.adoc
enikao: The problem is where the range starts and ends, especially when rendering shows more or less space between formats, or breaks the line inside the range.

dslmeinte: Could we shore up the notation, e.g. as follows: [_lower_.._upper_ ] ?

delta/messages.adoc
```diff
@@ -25,6 +25,9 @@ Can be `null`.
 Cannot be `null`.
 Use value `0` for <<{m3}.adoc#multiplicity, single links>>.
 
+[[indexOffsetType]]Index:: Relative position inside a multiple link or annotations as integer.
```
enikao: To readers, both types render as "Index" -- we should distinguish them.

delta/commandsEvents/children/MoveAndReplaceChildInSameContainment/ChildMovedAndReplacedInSameContainment.adoc
```diff
@@ -1,35 +1,35 @@
...
-<<existing-nodes, Existing node>> <<ChildMovedAndReplacedInSameContainment.movedChild>> (previously inside <<ChildMovedAndReplacedInSameContainment.parent>>'s <<ChildMovedAndReplacedInSameContainment.containment>> at <<ChildMovedAndReplacedInSameContainment.oldIndex>>) has replaced the <<existing-nodes, existing>> <<ChildMovedAndReplacedInSameContainment.replacedChild>> inside <<ChildMovedAndReplacedInSameContainment.parent>>'s <<ChildMovedAndReplacedInSameContainment.containment>> at <<ChildMovedAndReplacedInSameContainment.newIndex>>.
-<<ChildMovedAndReplacedInSameContainment.replacedChild>>, and all its <<ChildMovedAndReplacedInSameContainment.replacedDescendants>>{fn-org370}, have been deleted.
+<<existing-nodes, Existing node>> <<ChildMovedAndReplacedInSameContainment.movedChild>>
+that was initially at index <<ChildMovedAndReplacedInSameContainment.oldIndex>>
+in <<ChildMovedAndReplacedInSameContainment.parent>>'s annotations
```
enikao: "annotations" instead of containment

delta/commandsEvents/children/MoveAndReplaceChildInSameContainment/ChildMovedAndReplacedInSameContainment.adoc
```diff
@@ -1,35 +1,35 @@
...
+was moved by <<ChildMovedAndReplacedInSameContainment.indexOffset>> positions
+to index (<<ChildMovedAndReplacedInSameContainment.oldIndex>> + <<AnnotationMovedInSameParent.indexOffset>>),
```
enikao: "\<\<AnnotationMovedInSameParent.indexOffset>>" wrong reference

delta/commandsEvents/children/MoveAndReplaceChildInSameContainment/MoveAndReplaceChildInSameContainment.adoc
```diff
@@ -22,35 +19,36 @@ include::MoveAndReplaceChildInSameContainment-diagram.adoc[]
...
 
 .Errors
-* <<err-unknownIndex>> for <<MoveAndReplaceChildInSameContainment.newIndex>> (is beyond (the number of `<<MoveAndReplaceChildInSameContainment.movedChild>>'s containment entries in <<MoveAndReplaceChildInSameContainment.movedChild>>'s parent) - 2)
-* <<err-parentMismatch>> <<MoveAndReplaceChildInSameContainment.parent>> does not have <<MoveAndReplaceChildInSameContainment.movedChild>> as child.
-* <<err-unknownNode>> for <<MoveAndReplaceChildInSameContainment.movedChild>>
-* <<err-unknownNode>> for <<MoveAndReplaceChildInSameContainment.replacedChild>>
 * <<err-unknownNode>> for <<MoveAndReplaceChildInSameContainment.parent>>
-* <<err-parentMismatch>> The n ode <<MoveAndReplaceChildInSameContainment.parent>> does not have <<MoveAndReplaceChildInSameContainment.movedChild>> as a child.
+* <<err-unknownIndex>> for <<MoveAndReplaceChildInSameContainment.oldIndex>>
+is not an integer in the set [0..(the number of containment entries in <<MoveAndReplaceChildInSameContainment.containment>> of <<MoveAndReplaceChildInSameContainment.parent>> - 1)] — see <<integerRangeNotation>> for an explanation of the [`left`..`right`] notation
```
enikao: "set" or "range"?

delta/commandsEvents/children/MoveAndReplaceChildInSameContainment/MoveAndReplaceChildInSameContainment.adoc
```diff
@@ -22,35 +19,36 @@ include::MoveAndReplaceChildInSameContainment-diagram.adoc[]
...
+* <<err-invalidIndexOffset>> for <<MoveAndReplaceChildInSameContainment.indexOffset>>
+if <<MoveAndReplaceChildInSameContainment.indexOffset>> equals 0, or if (<<MoveAndReplaceChildInSameContainment.oldIndex>> + <<MoveAndReplaceChildInSameContainment.indexOffset>>) is not in the set [0..(the number of containment entries in <<MoveAndReplaceChildInSameContainment.containment>> of <<MoveAndReplaceChildInSameContainment.parent>> - 1)]
```
enikao: also set/range

delta/commandsEvents/children/MoveAndReplaceChildInSameContainment/MoveAndReplaceChildInSameContainment.adoc
```diff
@@ -22,35 +19,36 @@ include::MoveAndReplaceChildInSameContainment-diagram.adoc[]
...
+* <<err-unknownNode>> for <<MoveAndReplaceChildInSameContainment.replacedChild>>
+* <<err-parentMismatch>> <<MoveAndReplaceChildInSameContainment.parent>> is not the parent of <<MoveAndReplaceChildInSameContainment.replacedChild>>
+* <<err-indexEntryMismatch>> for <<MoveAndReplaceChildInSameContainment.containment>>/(<<MoveAndReplaceChildInSameContainment.oldIndex>> + <<MoveAndReplaceChildInSameContainment.indexOffset>>)
 and <<MoveAndReplaceChildInSameContainment.replacedChild>>
-* <<err-invalidMove>> as <<MoveAndReplaceChildInSameContainment.movedChild>> is already at <<MoveAndReplaceChildInSameContainment.newIndex>>
-* <<err-indexEntryMismatch>> for <<MoveAndReplaceChildInSameContainment.newIndex>> and <<MoveAndReplaceChildInSameContainment.replacedChild>>
-* <<err-indexEntryMismatch>> for <<MoveAndReplaceChildInSameContainment.oldIndex>> and <<MoveAndReplaceChildInSameContainment.movedChild>>
+if <<MoveAndReplaceChildInSameContainment.replacedChild>> is not at index (<<MoveChildInSameContainment.oldIndex>> + <<MoveChildInSameContainment.indexOffset>>) of the containment entries in <<MoveChildInSameContainment.containment>> of <<MoveAnnotationInSameParent.parent>>
```
enikao: Several references to `MoveChildInSameContainment`

delta/description.adoc
```diff
@@ -190,3 +190,124 @@ All values MUST be strings.
...
+For `indexOffset` > 0:
+[cols="a,a","a"]
+|===
+|Before |After
+include::moving/moving-and-replacing-diagram-offset-positive.adoc[]
+|===
+
+The new index of the moved child can be calculated as follows: `newIndex` = `oldIndex` + `indexOffset` - 1.
+
+Example moving `X` with `indexOffset` = 3, and replacing `F`: +
```
enikao, line 302: I'd also describe `targetIndex` explicitly in the examples.

delta/description.adoc
```diff
@@ -190,3 +190,124 @@ All values MUST be strings.
...
+1. Moving a node to a different position – the _target_ index – in a list, moving other list nodes aside as necessary.
+2. Moving a node to a different position – the _target_ index – in a list, *and* replacing the list node at that target index.
+Note that for this operation, the moved node might not reside at that target index afterward, because of the replacing!
+
+To avoid ambiguities in the description of such operations, and reduce the chance of misunderstanding the specification as much as possible, we chose to phrase such operations in terms of an _index offset_ — i.e., “in relative coordinates”.
+
+For convenience, we first establish some notation/shorthand:
+
+[[integerRangeNotation, notation/shorthand for integer ranges]]
+.Integer ranges
```
enikao, line 214: I find it confusing that we define several shorthands, but only one of them gets an intermediate heading. Proposal: Make the heading for notation/shorthand, and treat all shorthands uniformly (currently: bold / italics).

delta/description.adoc
```diff
@@ -190,3 +190,124 @@ All values MUST be strings.
...
+|===
+|Before |After
+include::moving/moving-diagram-offset-negative.adoc[]
+|===
+
+Example moving `X` with `indexOffset` = -2: +
+`oldIndex=3`: `A[0] B[1] C[2] *X[3]* D[4] E[5] F[6] G[7]` -> +
+`newIndex=1`: `A[0] *X[1]* _B[2] C[3]_ D[4] E[5] F[6] G[7]` +
+
+In both cases, only the nodes with initial indices in [**min**(`oldIndex`, `newIndex`)..**max**(`oldIndex`, `newIndex`)] are affected.
```
enikao: with the square brackets, I find the bold min/max less readable -- feels more like one of the notations ("node to move" etc.)

delta/description.adoc
```diff
@@ -190,3 +190,124 @@ All values MUST be strings.
...
+1. Set the node at index `targetIndex` to the node with initial index `oldIndex`.
+This action overwrites and effectively replaces the node to replace with the node to move.
+The node to replace gets deleted as a result.
+Now, the list contains the same number of nodes as before, just with two occurrences of the node to move, namely at indices `oldIndex` and `targetIndex`, and no occurrences of the node to replace.
+2. The node with index `oldIndex` is removed from the list, with the nodes after it being moved to the next lower position — this includes the node moved in step 1 itself if (`indexOffset` > 0).
+Now, the list contains one less node than before.
+
+For clarity, we discuss two distinct “sub” scenarios depending on whether `indexOffset` is positive or negative.
+
+For `indexOffset` > 0:
```
enikao, line 293: Shall we introduce sub-sections for the two cases? (either "=====Section" or ".Section")?

delta/description.adoc
```diff
@@ -190,3 +190,124 @@ All values MUST be strings.
...
+This pertains to the following commands and events:
+<<cmd-MoveChildInSameContainment>>, <<cmd-MoveAnnotationInSameParent>>, <<evnt-ChildMovedInSameContainment>>, <<evnt-AnnotationMovedInSameParent>>.
+
+Set `newIndex` = (`oldIndex` + `indexOffset`), which in this scenario is also the _target index_.
+The result of moving the node that’s initially at index `oldIndex` by `indexOffset` positions is that that node has been moved `indexOffset` positions up (if `indexOffset` > 0), or (-`indexOffset`) positions down (if `indexOffset` < 0).
+Afterward, the moved node resides at index `newIndex` in the list.
+
+For clarity, we discuss two distinct “sub” scenarios depending on whether `indexOffset` is positive or negative.
+
+For `indexOffset` > 0:
```
enikao, line 239: Maybe also introduce sub-sections?

delta/description.adoc
dslmeinte: Would unbolding min and max be enough?

delta/description.adoc
dslmeinte: I’m failing to see the problem here: the **node to move/replace** definitions are not notation/shorthand.

enikao: Different formats and paragraph styles for shorthand

The first paragraph is about "integer range" shorthand, it's formatted bold and has a sub-section header and a paragraph.

The next two sentences are about "old index" and "node to move" shorthand, they are formatted tt / italic / bold and don't have a paragraph.

The next two sentences are about "index offset" and "target index" shorthand, they are formatted tt / italic and dno't have a paragraph.

delta/commandsEvents/annotations/MoveAndReplaceAnnotationInSameParent/AnnotationMovedAndReplacedInSameParent.adoc
```diff
@@ -1,33 +1,29 @@
 [[evnt-AnnotationMovedAndReplacedInSameParent]]
 ===== Annotation moved in same parent and replaced existing annotation
-<<existing-nodes, Existing node>> <<AnnotationMovedAndReplacedInSameParent.movedAnnotation>> (previously inside <<AnnotationMovedAndReplacedInSameParent.parent>>'s annotations at <<AnnotationMovedAndReplacedInSameParent.oldIndex>>) has replaced the <<existing-nodes, existing>> <<AnnotationMovedAndReplacedInSameParent.replacedAnnotation>> inside <<AnnotationMovedAndReplacedInSameParent.parent>>'s annotations at <<AnnotationMovedAndReplacedInSameParent.newIndex>>.
-<<AnnotationMovedAndReplacedInSameParent.replacedAnnotation>>, and all its <<AnnotationMovedAndReplacedInSameParent.replacedDescendants>>{fn-org370}, have been deleted.
+<<existing-nodes, Existing node>> <<AnnotationMovedAndReplacedInSameParent.movedAnnotation>>
+that was initially at index <<AnnotationMovedAndReplacedInSameParent.oldIndex>>
+in <<AnnotationMovedAndReplacedInSameParent.parent>>'s annotations
+was moved by <<AnnotationMovedAndReplacedInSameParent.indexOffset>> positions
+to index (<<AnnotationMovedAndReplacedInSameParent.oldIndex>> + <<AnnotationMovedInSameParent.indexOffset>>),
```
enikao: "AnnotationMovedInSameParent.indexOffset" wrong reference

delta/commandsEvents/annotations/MoveAndReplaceAnnotationInSameParent/MoveAndReplaceAnnotationInSameParent.adoc
```diff
@@ -1,20 +1,15 @@
 [[cmd-MoveAndReplaceAnnotationInSameParent]]
 ===== Move annotation in same parent and replace existing annotation
-Move <<existing-nodes, existing node>> <<MoveAndReplaceAnnotationInSameParent.movedAnnotation>>
-from <<MoveAndReplaceAnnotationInSameParent.oldIndex>>
-to <<MoveAndReplaceAnnotationInSameParent.newIndex>>.
-within <<MoveAndReplaceAnnotationInSameParent.parent>>'s annotations.
-
-Delete <<existing-nodes, current node>> <<MoveAndReplaceAnnotationInSameParent.replacedAnnotation>>{fn-org371} inside <<MoveAndReplaceAnnotationInSameParent.parent>>'s
-annotations at <<MoveAndReplaceAnnotationInSameParent.newIndex>>, and all its descendants (including annotation instances).
+Move <<existing-nodes, existing node>> <<MoveAnnotationInSameParent.movedAnnotation>>
```
enikao: "MoveAnnotationInSameParent" several wrong references

enikao (commented):
we're getting there (-:

delta/description.adoc
dslmeinte: Provided a header for the 2nd paragraph. For the rest, I don’t really want to change too much anymore.

enikao (approved):

PR #496, opened 2026-07-08
Spell "data type" correctly

dslmeinte:

PR #497, opened 2026-07-15
Missed two misspelled "datatype"s

dslmeinte:
(Well, the same one really.)
Not sure we want to change the (published) 2024.1 spec.

### DIFF
--- a/2024.1/serialization/serialization.adoc
+++ b/2024.1/serialization/serialization.adoc
@@ -287,7 +287,8 @@ Booleans MUST NOT be encoded with leading or trailing whitespace, uppercase char
 * `"-6 "`
 
 [[structuredDatatype]]
-==== Structured Datatype
+[[structuredDataType]]
+==== Structured Data Type
 <<{m3}.adoc#StructuredDataType, LionCore StructuredDataType>> MUST be encoded as JSON _string_.
 The string contains a JSON _object_ according to spec (RFC 8259) with proper escaping: all double quotes, line breaks, etc. MUST be escaped to form a proper JSON _string_.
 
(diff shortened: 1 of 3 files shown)


2024.1/serialization/serialization.adoc
```diff
@@ -287,7 +287,8 @@ Booleans MUST NOT be encoded with leading or trailing whitespace, uppercase char
 * `"-6 "`
 
 [[structuredDatatype]]
-==== Structured Datatype
+[[structuredDataType]]
```
enikao: I would not change a published spec.
If we want to, we can add it to the errata: https://lionweb.io/specification/2024.1/index.html

dslmeinte: I feel it’s a bit...trivial for the errata, but can do for those with OCD 😏

enikao: I agree with the triviality (-:

2024.1/serialization/serialization.adoc
```diff
@@ -297,7 +297,7 @@ Each <<{m3}.adoc#Field, Field>> forms one _member_, with the field's <<{m3}.adoc
 For fields of type <<string>>, <<boolean>>, <<integer>>, and <<literal>>, the _value_ is encoded as JSON _string_ in the same way as for a property.
 For fields of type <<structuredDatatype>>, the value is encoded as JSON _object_.
 
-.Structured data types used in the examples
+.Structured datatypes used in the examples
```
enikao, line 300: but it is changed now?

dslmeinte: I noticed that I had changed it already earlier (in PR #496), so I reverted those changes in the latest commit.

enikao (approved):

Issue #438, opened 2026-02-11
UnsubscribeFromChangingPartitions


joswarmer:
For `SubscribeToPartitionContents` we have `UnsubscribeFromPartitionContents` to unsubscribe.

For `SubscribeToChangingPartitions` and `InformAboutChangingPartitions` we don't have an unsubscribe.
I think we should have a `UnsubscribeFromChangingPartitions` to cancel `SubscribeToChangingPartitions` and `InformAboutChangingPartitions`.

joswarmer:
In that case we need an "unsubscribe" request for both of these, otherwise a client cannot change it partition subscription type.

enikao:
> In that case we need an "unsubscribe" request for both of these, otherwise a client cannot change it partition subscription type.

We have that -- send the same query with `added` / `deleted` = false

joswarmer:
> > In that case we need an "unsubscribe" request for both of these, otherwise a client cannot change it partition subscription type.
> 
> We have that -- send the same query with `added` / `deleted` = false

I didn't realize that you could do this, nice!

PR #499, opened 2026-07-22
added details about subscribe vs. inform partition changes queries

enikao:


dslmeinte (changes requested):
I’m missing the “How to distinguish whether an PartitionAdded event contains the whole partition” part of #438 being propagated to this PR.

dslmeinte (approved):

Issue #352, opened 2025-03-23
Can a client change a node it is not subscribed to?


enikao:
Assume 
* client `A` did subscribe to `partitionX`, but NOT to `partitionY`
* `partitionX` contains `nodeX`
* `partitionY` contains `nodeY`
* `nodeY` contains `nodeYChild`

Can client `A`:
1. Change anything about `nodeY`?
2. Move `nodeYChild` under `nodeX`?

enikao:
Makes sense, but how would the client ever know whether the change succeeded?

joswarmer:
E.g. the following scenario:
- the repository supports locking partitions. 
- The client locks one partition to edit it. 
- The client subscribes to changes from all other partitions.
The client now assumes (at its own risk) that his commands will be executed as there can be no merge conflicts.

joswarmer:
We can also say that a client will always gets events for changes that he initiated (i.e. his command id is in the event), even when the client is not subscribed to them.
That will break however, if a client has zero subscriptions, as the server would not know how to send the event.

PR #505, opened 2026-07-30
clarify that a client always gets the events of its own commands

enikao:
based on #352

### DIFF
--- a/delta/description.adoc
+++ b/delta/description.adoc
@@ -111,7 +111,7 @@ Once all <<qry-InformAboutChangingPartitions>> are disabled (i.e. both <<InformA
 A client CAN send commands for any node id, independent of the client's subscription.
 Assume a client sends a command for a node that's part of a partition that client is not subscribed to.
 The repository treats that command the same as from a client that was subscribed to the containing partition, i.e. applies it in accordance with the repository's policy.
-The repository sends out the same events to subscribed clients, _including_ to the sending client.{fn-org352}
+The repository sends out the same events to subscribed clients, and _also_ to the sending client.{fn-org352}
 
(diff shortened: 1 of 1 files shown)


delta/description.adoc
```diff
@@ -108,10 +108,15 @@ However, a client MUST NOT mix <<qry-InformAboutChangingPartitions>> and <<qry-S
 If any <<qry-InformAboutChangingPartitions>> query is active, the same client MUST NOT issue a <<qry-SubscribeToChangingPartitions>> query, and vice versa.
 Once all <<qry-InformAboutChangingPartitions>> are disabled (i.e. both <<InformAboutChangingPartitions.creation>> and <<InformAboutChangingPartitions.deletion>> are `false`), the client MAY issue a <<qry-SubscribeToChangingPartitions>> query.{fn-org438}
 
-A client can send commands for any node id, independent of the client's subscription.
+A client CAN send commands for any node id, independent of the client's subscription.
 Assume a client sends a command for a node that's part of a partition that client is not subscribed to.
 The repository treats that command the same as from a client that was subscribed to the containing partition, i.e. applies it in accordance with the repository's policy.
-The repository also sends out the same events to subscribed clients, thus _not_ to the sending client.{fn-org352}
+The repository sends out the same events to subscribed clients, _including_ to the sending client.{fn-org352}
```
dslmeinte: I find “including” a bit misleading here: if the client is _not_ subscribed to (changes on) the node it sends a command for, then the repository has to send the corresponding event _also_ to the sending client.

enikao: fixed

dslmeinte (approved):

